Issue #15, opened 2015-07-24
[BUG] self.stopSimulation() doesn't call steppable.finish()


zafarali:
Hi,

I noticed that calling `self.stopSimulation()` in a steppable does not call the `finish()` for the registered steppables. Is this a deliberate design decision? I would assume that `finish()` runs regardless of how the simulation has been ended.

tjsego:
@CompuCell3D this is resolved by #287. Can we close this?

PR #16, opened 2015-07-24
stopSimulation() fix

zafarali:
- Refer bug in #15, this patch will call the `finish()` methods of all the registered steppables.

Note: Not sure if this was a deliberate design decision and I might have missed something, but I think there should be a way to call the `finish()` methods either by passing in another argument or by default.


CompuCell3D:
We are moving to CC3D 4.0.0 and the functionality in this PR is already in 4.0.0 branch. Closing the PR without merge for now. Thank you for pointing out this issue

Issue #18, opened 2015-11-27
PIFDumper steppable ignores Frequency that is inserted by Twedit


burgerga:
Twedit generates:
    
    <Steppable Frequency="100" Type="PIFDumper">
        
        <PIFName>NewSimulation</PIFName>
        <PIFFileExtension>piff</PIFFileExtension>
    </Steppable>

but in PIFDumper.cpp the Frequency is ignored.

maciekswat:
Hi Gerhard

I ran the simulation with frequency set to 10 or a 100 and it appears to work correctly. Which version of CC3D are you using? I noticed though that the output is written to the directory in which CC3D is installed, which is not the best solution so I need to add code to write to simulation output directory. Let me know if you still have troubles with PIFDumper. BTW you can write PIF's directly from Python if you are interested in this I can send you code snippet that does it.

burgerga:
You're right, it works correctly. I wanted to write my own steppable with a Frequency argument and based on

    void PIFDumper::update(CC3DXMLElement *_xmlData, bool _fullInitFlag){
    
       //Check how this is handled 
	    //frequency=frequency;
	    pifname=_xmlData->getFirstElement("PIFName")->getText();
	    if(_xmlData->findElement("PIFFileExtension"))
	    	pifFileExtension=_xmlData->getFirstElement("PIFFileExtension")->getText();

    	//frequency=1;
    }

I thought that the frequency was ignored. So I guess it must be handled at a higher level.
Sorry for the confusion.

burgerga:
As for the output in the install directory, I fixed it in c4f0f610024ca76e42d00c4651d41afbc87f9d74 in my fork, I will prepare a pull request soon.

maciekswat:
I saw your fix and I initially came up with the same solution but this will write the output to the directory where the simulation scripts are. I think it would be better to have a code that writes output to the simulation output directory and there are few CC3D c++ steppables that would benefit from this. In any case as soon as I find some time I will fix it. I know what needs to be done. I might reuse part of your pull request. If you would like to try implementing the feature I think you would need to add getter/setter methods to Simulator class that would allow setting and reading the location of the output directory. The actual directory is created in Python I believe... The issue might be what happens if user does not request any output from the Player or runScript. I guess I will need to change the way we handle creation of the output directory

PR #19, opened 2015-11-30
Fix PIFDumper output directory 

burgerga:
Would previously output piff files in the installation directory instead of simulation directory as mentioned in https://github.com/CompuCell3D/CompuCell3D/issues/18.


CompuCell3D:
As far as the path we decided that user should specify full output path in the PIFDumper. We are closing this PR without merge though. 
we are moving to Python 3- based  CC3D 4.0.0 and all future PRs shold go there. thank you for pointing out the issue

PR #32, opened 2016-11-23
Fixes a NameError bug

IsmailM:
When using the Chemotaxis Plugin where cells undergo mitosis, you see the following error:

   NameError: global name 'chemotaxisPlugin' is not defined

This fixes it.


CompuCell3D:
Hi Ismail, thanks for catching it

PR #109, opened 2018-12-21
fix cloning of chemotaxis attributes

VoisinneG:


CompuCell3D:
Great. thank you for catching it!

PR #110, opened 2019-02-12
Neighbour surface constraint

JulianoGianlupi:


JulianoGianlupi:
did it by accident, disregard

PR #111, opened 2019-02-12
Demo fixing

JulianoGianlupi:


jpsluka:
Juliano-

 

Could you replace this demo with the version attached to this email in the new Demos directory? The old version wrote 10000 txt files (each 255KB) into the Compucell3d directory. This version writes to the projects output dir and only generate 10 files.

 

This goes in the demos directory:

…\Demos\SteppableDemos\secretion_total_count\

(Note that the CC3D job file names don’t match the directory name)

 

jim

JulianoGianlupi:
Sure thing, only I didn't get the attachment.

Also, I believe I created this pull request by accident while merging stuff in my own fork. There are a few more fixes that I may be able to do before pulling into the base.

Juliano

JulianoGianlupi:
I'll create a new pull with the most recent changes

PR #167, opened 2019-11-16
Fixed no update of screen update frequency from GUI preferences

tjsego:
GUI fix: changing screen update frequency in preferences had no effect. Simple method added to SimulationThread updates corresponding member from configuration database if accessible. Call the method before resuming simulation.

### DIFF
--- a/cc3d/CompuCellSetup/simulation_setup.py
+++ b/cc3d/CompuCellSetup/simulation_setup.py
@@ -556,6 +556,8 @@ def main_loop_player(sim, simthread=None, steppable_registry=None):
     if init_using_restart_snapshot_enabled:
         steppable_registry.restart_steering_panel()
 
+    simthread.setScreenUpdateFrequency()
+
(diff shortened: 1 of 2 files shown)


CompuCell3D:
HI TJ, so I checked on Windows and changing screen update frequency works with latest master snapshot. What were your other settings when you observed that it does not function properly? I mean did you set screenshot output frequency? And which version of CC#D did you run? was it master branch of 4.1.0?

CompuCell3D:
HI TJ, so I checked on Windows and changing screen update frequency works with latest master snapshot. What were your other settings when you observed that it does not function properly? I mean did you set screenshot output frequency? And which version of CC#D did you run? was it master branch of 4.1.0?

tjsego:
Hi Maciek. I've observed this for a number of different simulations on my machine using Windows 8.1 with the latest release and the master branch (as of this morning).

CompuCell3D:
Merged

PR #168, opened 2019-11-18
Added command line argument for setting vtk core names during output

tjsego:
Returned command line argument to modify vtk output file core name.


CompuCell3D:
Merged

PR #176, opened 2019-12-07
fixed workspace dir option in the Preferences - previously even if us…

CompuCell3D:
…ers set it to some directory of their choice it was "hard-wired" to return <home_dir>/CC3DWorkspace. This commit fixes this bug

### DIFF
--- a/cc3d/CompuCellSetup/persistent_globals.py
+++ b/cc3d/CompuCellSetup/persistent_globals.py
@@ -43,6 +43,8 @@ class PersistentGlobals:
         self.__output_dir = None
         self.output_file_core_name = "Step"
 
+        self.__workspace_dir = None
+
         self.output_frequency = 0
         self.screenshot_output_frequency = 0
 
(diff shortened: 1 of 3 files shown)


cc3d/CompuCellSetup/persistent_globals.py
```diff
@@ -96,18 +98,32 @@ def set_output_dir(self, output_dir: str) -> None:
...
-            Path(workspace_dir).mkdir(parents=True, exist_ok=True)
+        if self.__workspace_dir is None:
+            workspace_dir = os.path.join(os.path.expanduser('~'), 'CC3DWorkspace')
+            if not exists(workspace_dir):
+                Path(workspace_dir).mkdir(parents=True, exist_ok=True)
 
-        return workspace_dir
+            return workspace_dir
+        else:
+            return self.__workspace_dir
```
JulianoGianlupi: Wouldn't it be interesting and user friendly to make the user inputed directory if it doesn't exist?

tjsego: Agreed.

CompuCell3D: It is already implemented

PR #183, opened 2020-01-07
Feature/fpp initiator tracking

tjsego:
Added new boolean to focal point plasticity tracker data for which cell initiated a link, plus improvements to python plugin API and a demo.


CompuCell3D:
looks ok. Approving it

tjsego:
> This looks OK but I am not sure why we need to know which function "initiated" the link. Link gets established if two cells come in contact so the question which cell initiated is not really that meaningful in Potts. But the code looks fine. I mean it is not a breaking change

It's a requested feature. I'm not exactly clear on its usage.

Issue #161, opened 2019-10-05
Missing  menu item in Twedit to turn off insertion of comments in snippets


maciekswat:

tjsego:
Fixed by https://github.com/CompuCell3D/CompuCell3D/pull/185

PR #185, opened 2020-01-10
Fix to missing optional snippet comments

tjsego:
Concerning https://github.com/CompuCell3D/CompuCell3D/issues/161
Fixed issue prohibiting python menu, plus implemented same approach to XML
C++ snippets don't have many comments, so not implemented here.


CompuCell3D:
Approving it

PR #196, opened 2020-02-04
Feature/osx layout screenshots

CompuCell3D:
This implements rudimentary screenshot description file viewer + adds new API to the PySteppableBase that allows taking screenshots at arbitrary mcs.


CompuCell3D:
The question now is whether we also want to specify custom directory for those ad-hoc screenshots. Right now they are being written to the same directory as other screenshots

tjsego:
> The question now is whether we also want to specify custom directory for those ad-hoc screenshots. Right now they are being written to the same directory as other screenshots

I think keeping ad-hoc screenshots exported to the same directory as other screenshots is much cleaner.

PR #207, opened 2020-02-16
Changed plot "add data point" snipet 

JulianoGianlupi:
Now it is 

```
self.plot_win.add_data_point("MVol", mcs, cell.volume)
self.plot_win.add_data_point("MSur", mcs, cell.surface)
```

instead of 

```
self.plot_win.add_data_point("MVol", mcs, mcs)
self.plot_win.add_data_point("MSur", mcs, mcs**2
```


CompuCell3D:
NIce

PR #209, opened 2020-02-18
Added template simulation files to demos for better accessibility

tjsego:
In support of model building from demos, per request from JAG. 
Template files are per current Wizard templates.


CompuCell3D:
TJ, Just to make sure you are adding steppables to simulations that were previously XML only?

tjsego:
> TJ, Just to make sure you are adding steppables to simulations that were previously XML only?

Correct. This was just so that model building can begin from any available demo without needing to add/set up any files.

CompuCell3D:
The downside is that you are calling empty functions in steppable. I guess this is not too much of a burden

PR #211, opened 2020-02-20
Added energy calculations in Python

tjsego:
Per request from J. Sluka


tjsego:
> This looks good but Could you also add example of simulation that uses it. I guess you need to specify in the XML that you want to keep track of calculated energies and then use Python API to access those So the process of setting it up is not trivial and example would help a lot

See https://github.com/CompuCell3D/CompuCell3D/pull/211/commits/cb7d1eb885782ed435fc6777bd46d323e39e2502. Let me know if this is sufficient.

CompuCell3D:
Sure this looks fine

PR #213, opened 2020-02-20
Volume parameters initialized from XML if not specified in steppable

tjsego:
Per JAG's request: CellG instance members `targetVolume` and `lambdaVolume` should be initialized based on XML. To accommodate setting in steppable `start` methods, I added on-the-fly initialization at first call to `changeEnergy` to populate instances with values according to whatever is specified in XML by type (if anything). So the precedence is `steppable.start() assignments by cell > XML spec by type > default = 0` for both members. 

The only limitation is that if nonzero member values are specified in XML for a type and both are set to zero during steppable `start` for a CellG instance of that type, the members of the instance will be overwritten according to the XML during the first step. However, this seems like an unlikely scenario.

### DIFF
--- a/CompuCell3D/core/CompuCell3D/plugins/Volume/VolumePlugin.cpp
+++ b/CompuCell3D/core/CompuCell3D/plugins/Volume/VolumePlugin.cpp
@@ -183,6 +183,54 @@ void VolumePlugin::extraInit(Simulator *simulator){
 	update(xmlData);
 }
 
+void VolumePlugin::initializeCellVolumeParams() {
+
+	if (cellVolumeParamsInitialized) { return; }
+
+	// Assign whatever cell type parameters are specified in XML to all cells by type, which may or may not be overwritten by steppables during start()
+
+	CC3DXMLElementList XMLParamsList = xmlData->getElements("VolumeEnergyParameters");
+	CellInventory *cellInventoryPtr = &potts->getCellInventory();
+	CellInventory::cellInventoryIterator cInvItr;
+	Automaton *automaton = potts->getAutomaton();
+	std::map<unsigned char, std::tuple<float, float> > typeParamsMap;
+
+	for each (CC3DXMLElement *XMLParams in XMLParamsList) {
+
+		std::string cellTypeName = XMLParams->getAttribute("CellType");
+
+		try {
+
+			unsigned char typeId = automaton->getTypeId(cellTypeName);
+			float targetVolume = (float)XMLParams->getAttributeAsDouble("TargetVolume");
+			float lambdaVolume = (float)XMLParams->getAttributeAsDouble("LambdaVolume");
+
+			typeParamsMap.insert(make_pair(typeId, make_tuple(targetVolume, lambdaVolume)));
+
+		}
+		catch (BasicException &e) { cerr << "Unrecognized cell type: " << cellTypeName << endl; }
+
+	}
+
+	std::map<unsigned char, std::tuple<float, float> >::iterator mapItr;
+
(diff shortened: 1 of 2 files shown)


CompuCell3D:
I am not sure if this makes sense. What if you create new cells in the simulation. Will you be also initializing targetVolume and lambdaVolume in this situation? 
Another issue, even if you do initialization what does it buy you? If you are using global settings from XML then by definition you will not be calling energy function that uses cell->targetVolume. In fact you confuse users who would think that volume plugin is indeed using cell->targetVolume. 

Another point what if I change XML while simulation runs (known as steering) To be consistent I would also need to rerun  initialize_target_volume. 

If I were a user and I saw cell->targetVolume=28 but plugin calculated energies as if targetVolume was 15 then I would be confused and I can easily create such situation with CC3D. I really think it is better to keep current code and not introduce extra logic that is hard-to-maintain

JulianoGianlupi:
I agree with Maciej. This seems to try to solve a non existing issue and will more than likely cause head aches for users and developers

JulianoGianlupi:
I was thinking the past few days, why do we even have global, cell type, cell id as different behaviors in c++?

We could read from XML and write into the cell object instead of keeping a list. This way we only need one energy function and the attributes will always be available in python without any xml-reading function.

How does this sound? I don't see this breaking anything but I could be wrong

tjsego:
> I was thinking the past few days, why do we even have global, cell type, cell id as different behaviors in c++?
> 
> We could read from XML and write into the cell object instead of keeping a list. This way we only need one energy function and the attributes will always be available in python without any xml-reading function.
> 
> How does this sound? I don't see this breaking anything but I could be wrong

If I've understood this correctly, then there would only be the cell data parameters for constraint and coefficient, and no type-specific list maintained by the plugin. Is that correct?

If so, I'm inclined to agree in the sense that computations should be done by accessing parameters by cell, rather than by using cell type as a key to a global list. In my opinion, much of the strength of the framework is being able to edit parameters by cell in Python. I would only keep the parameter data by type in the plugin as a reference for memory purposes (not computer memory) so that, say, a cell's parameters could be reset to the defined parameters by cell type on-the-fly. 

But in general, I think that parameter operations should be done at the level of the cell, and not the type. Granted, it requires more memory, but I think it's worth it.

CompuCell3D:
Juliano, we have those functions for compatibility reasons. At some point we could break compatibility if we have good reasons. I think that given our backlog we could postpone further discussion of volume plugin . I really do not think this is that important. I think what would be more important is to think about more comprehensive overhaul of our specifications. I am in favor of per-cell definitions of parameters . This would work in some cases but keep in mind that we also have interactions and those are best recorded in some type of "global" object. Anyway as far as this particular PR I suggest we do not include it . If we want to do something radical we need to do it in a coordinated effort and address issues like that for all plugins. What do you think?

tjsego:
> Juliano, we have those functions for compatibility reasons. At some point we could break compatibility if we have good reasons. I think that given our backlog we could postpone further discussion of volume plugin . I really do not think this is that important. I think what would be more important is to think about more comprehensive overhaul of our specifications. I am in favor of per-cell definitions of parameters . This would work in some cases but keep in mind that we also have interactions and those are best recorded in some type of "global" object. Anyway as far as this particular PR I suggest we do not include it . If we want to do something radical we need to do it in a coordinated effort and address issues like that for all plugins. What do you think?

Agreed.

JulianoGianlupi:
Agreed

CompuCell3D:
is it OK to close this PR?

tjsego:
Works for me.

JulianoGianlupi:
Me too

PR #215, opened 2020-02-24
Improvements to DiffusionSolverFE and ReactionDiffusionSolverFE

tjsego:
No changes to actual calculations. Just restructuring to remove redundant loops and calculations.

### DIFF
--- a/CompuCell3D/core/CompuCell3D/steppables/PDESolvers/DiffusionSolverFE.cpp
+++ b/CompuCell3D/core/CompuCell3D/steppables/PDESolvers/DiffusionSolverFE.cpp
@@ -835,101 +835,63 @@ void DiffusionSolverFE<Cruncher>::prepCellTypeField(int idx){
     // cerr<<"fieldDim="<<fieldDim<<endl;
     
     for (int iter = 0 ; iter < numberOfIters ; ++iter){
+
+		int x1Set = 0;
+		int x2Set = fieldDim.x + 1;
+		int x1Get, x2Get;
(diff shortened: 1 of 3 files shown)


tjsego:
I'll accomplish any other improvements in the [ECM_model](https://github.com/CompuCell3D/CompuCell3D/tree/ECM_model) branch.

tjsego:
> Looks good. Just as a matter of curiosity, what's the speedup?

For a 500 x 500 lattice, 2 fields, no cells, 4 solver iterations per step, Dirichlet boundary conditions, 8 cores, no rendering or reaction terms, and 10k steps, I measured about a 15% speedup for DiffusionSolverFE, and about 23% for ReactionDiffusionSolverFE.

CompuCell3D:
I think we might actually do it now an write regression testing for PDE solvers. This way we would feel much more comfortable approving changes like this one. TJ, can I ask you to write code that does comparison of PDE Solver outputs?

tjsego:
> I think we might actually do it now an write regression testing for PDE solvers. This way we would feel much more comfortable approving changes like this one. TJ, can I ask you to write code that does comparison of PDE Solver outputs?

Definitely. We should have a test suite for future development, anyway.

tjsego:
> There are actual changes in the output due to new code. I issued new PR that adds regresion tets for few PDE solvers and based on this comparison I saw some changes. Let's reconsile those offline

Sounds good. See [PDE_test_suite](https://github.com/CompuCell3D/CompuCell3D/tree/PDE_test_suite). I've already been putting some effort towards doing this, too.

PR #217, opened 2020-02-26
Fixed incompatible argument type in window position suggestions

tjsego:


jpsluka:
I've tried this patch on my Mac (OS 10.15.3) and it appears to fix the problems. Can open new windows and change what is displayed in a window without crashing Player.

PR #221, opened 2020-03-24
Added demo for screenshot API

CompuCell3D:
fixing bug in UserInterface.py - Redirector was behaving weirdly when one tried to visualize field
This fixes crashes during chemical  field visualization


jpsluka:
Any idea how the error was propagating back to previous version of CC3D (4.1.0 and 4.0.0)? Was the error leaving something in the registry or not closing a file correctly?

PR #226, opened 2020-04-01
Fluctuation compensator

tjsego:
Add-on feature for DiffusionSolverFE and ReactionDiffusionSolverFE: mass conservation in each subdomain (cells and medium) is enforced over all spin flips of a MCS. I.e., no more leaky cell boundaries. 

Be advised: this includes (minor) core mods.


CompuCell3D/core/CompuCell3D/steppables/PDESolvers/FluctuationCompensator.h
```diff
@@ -0,0 +1,161 @@
+/*************************************************************************
+*    CompuCell - A software framework for multimodel simulations of     *
+* biocomplexity problems Copyright (C) 2003 University of Notre Dame,   *
```
JulianoGianlupi, line 3: maybe cr should be for 2020 IU. I think new software modules can get a "current year" cr in the place they were made

PR #232, opened 2020-04-19
Added Player resize callback override to handle new pref. storage

tjsego:


CompuCell3D:
I think I will skip this PR. I implemented a fix that implements required logic regarding window resizing policy inside initialize_gui_geometry() function of the UserInterface.py

Issue #242, opened 2020-05-08
Add calculator of seen field by cell


JulianoGianlupi:
Basically the result of the uptake function without actually doing the uptake

CompuCell3D:
good feature

Issue #240, opened 2020-05-08
Report total of chemical fields


JulianoGianlupi:
Add to the solvers a tracker of the total value of a field (integral of the field).

The solvers already perform this calculation, it's a matter of finding a way to keep track of it

CompuCell3D:
could be good for testing as well

PR #243, opened 2020-05-12
Added convenience functions to FieldSecretor to compute 1) field "see…

CompuCell3D:
1)field "seen" by cell, 2) total field integral. This addresses

https://github.com/CompuCell3D/CompuCell3D/issues/242
https://github.com/CompuCell3D/CompuCell3D/issues/240

### DIFF
--- a/CompuCell3D/core/CompuCell3D/plugins/Secretion/FieldSecretor.cpp
+++ b/CompuCell3D/core/CompuCell3D/plugins/Secretion/FieldSecretor.cpp
@@ -772,3 +772,37 @@ bool FieldSecretor::uptakeInsideCellAtCOM(CellG * _cell, float _maxUptake, float
 
 
 }
+
+float FieldSecretor::_amountSeenByCell(CellG * _cell) {
+
+    if (!pixelTrackerPlugin) {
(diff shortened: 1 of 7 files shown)


CompuCell3D/core/CompuCell3D/plugins/Secretion/FieldSecretor.cpp
```diff
@@ -772,3 +772,37 @@ bool FieldSecretor::uptakeInsideCellAtCOM(CellG * _cell, float _maxUptake, float
 
 
 }
+
+float FieldSecretor::_amountSeenByCell(CellG * _cell) {
+
+    if (!pixelTrackerPlugin) {
+        // ASSERT_OR_THROW("PixelTracker Plugin has been turned off. Cannot execute _amountSeenByCell function",pixelTrackerPlugin);
+        
+        return -1.0;
```
JulianoGianlupi, line 781: Can't users encounter this and become confused?

CompuCell3D: This is handled in SWIG file. But  good point, perhaps we should also keep it here if somebody decides to use if from a steppable written in C++. For, now i think it is safe to assume it will be called via python interface though

PR #263, opened 2020-06-20
Fixed lost secretion in array swapping for ReactionDiffusionSolverFE

tjsego:


tjsego:
Current implementation in the solver prohibits usage without scaling, in which case diffusion is first executed, and then secretion. Before this fix, the solver places diffusion calculations in t+1, secretion terms in t, and then implements the array swap. This is how secretion is currently getting lost. The fix places secretion updates into t+1 _before_ the array swap, producing expected behavior in all of my tests.

tjsego:
This simple modification can do the trick. The algorithm is something like
Secretion: getDirect->setDirect
Diffusion/Reaction: getDirect->setDirectSwap
Swap arrays

CompuCell3D:
So previously the order of operations was 
1) diffusion
2) secretion

and with the fix is

1) secretion
2) diffusion

am I getting this right? If so then indeed the fix is the right way of handling secretion.

CompuCell3D/core/CompuCell3D/steppables/PDESolvers/ReactionDiffusionSolverFE.cpp
```diff
@@ -813,11 +813,11 @@ void ReactionDiffusionSolverFE::step(const unsigned int _currentStep) {
             for (int idx = 0; idx < numberOfFields; ++idx) {
                 prepCellTypeField(idx); // here we initialize celltype array  boundaries - we do it once per  MCS
                 boundaryConditionInit(idx);//initializing boundary conditions
-                solveRDEquationsSingleField(idx); //reaction-diffusion
                 //secretion
                 for (unsigned int j = 0; j < diffSecrFieldTuppleVec[idx].secrData.secretionFcnPtrVec.size(); ++j) {
                     (this->*diffSecrFieldTuppleVec[idx].secrData.secretionFcnPtrVec[j])(idx);
                 }
+				solveRDEquationsSingleField(idx); //reaction-diffusion
```
CompuCell3D, line 820: OK, one more request/ question to consider. Shouldn't we firs do secretion in all fields and then do the loop to solveRD equations? Something like this:

for(all fields):
   perform secretion

for (all fields)
   solveRDEquationSingleField

I think this would be correct order of operations . 

Otherwise during solveRDEquationsSingleField call for some fields we would be using field values with secretion accounted for and some field values would be using values before secretion took place

tjsego: Agreed.

CompuCell3D:
Ready to merge

Issue #253, opened 2020-06-02
Can Player's zip-it guess the zip files name?


jpsluka:
Can Player's zip-it guess the zip files name? Currently the file dialog comes up without any guess at what the zip file should be named. Can it just guess "TheProjectName.zip"?

CompuCell3D:
PR issued

CompuCell3D:
Fixed

PR #266, opened 2020-06-24
Feature/twedit zip it fix

CompuCell3D:
Most of changes are related to replacing OSX line endings with Unix ones so that github can display diffs correctly. NO need to review those changes. The only file that has non-trivial changes and implements Fix to https://github.com/CompuCell3D/CompuCell3D/issues/253 is this one:

cc3d/twedit5/Plugins/PluginCCDProject.py


CompuCell3D:
Merging PR.

Issue #259, opened 2020-06-13
on-demand screenshots API does not respect TypeColorMap property of screenshots.json file


gabroo:
changing colors in this map does not change the colors in resultant screenshots

CompuCell3D:
Could you post a simple simulation showing this issue along wtih json file?

CompuCell3D:
To do - store lookup tables for screenshots to avoid generating it each time screenshot is taken

CompuCell3D:
Fixed

PR #267, opened 2020-06-27
Fixing https://github.com/CompuCell3D/CompuCell3D/issues/259. Also di…

CompuCell3D:
…d some additional minor code cleanup

Fixed issue that screenshot generation routines did not respect the colors stored in screenshot description file but instead used whatever Player settings were at the moment

### DIFF
--- a/cc3d/core/GraphicsOffScreen/MVCDrawModelBase.py
+++ b/cc3d/core/GraphicsOffScreen/MVCDrawModelBase.py
@@ -28,22 +28,9 @@ class MVCDrawModelBase:
 
         self.celltypeLUT = None
 
-    #        self.scaleGlyphsByVolume = False
-
     # should also set "periodic" boundary condition flag(s) (e.g. for drawing FPP links that wraparound)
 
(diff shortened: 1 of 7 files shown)


cc3d/core/GraphicsOffScreen/MVCDrawModelBase.py
```diff
@@ -61,56 +48,64 @@ def setParams(self):
         self.lutBlueRed.Build()
 
     def get_type_lookup_table(self, scene_metadata=None):
+        try:
+            actual_screenshot = scene_metadata['actual_screenshot']
+        except KeyError:
```
tjsego, line 53: Think this needs to handle TypeError as well, for when scene_metadata is None

CompuCell3D: Good point

CompuCell3D: Fixed

PR #277, opened 2020-07-22
Fix/tracking

tjsego:
Fixes two issues in GUI-less mode

1. Extra fields weren't being written to file
2. start was called once in main loop when not restarting, and then called again whether beginning or restarting a simulation. Along with not being appropriate for restarting a simulation, this can potentially cause an error depending on user-defined steppables (_e.g._, calling start twice that creates an extra field when running a new simulation)


CompuCell3D:
Should be ready to merge

PR #287, opened 2020-07-28
Feature/on stop

CompuCell3D:
Implementing `on_stop` function in the steppable to be called when user stops simulation


CompuCell3D:
updated twedit++ templates as well

PR #336, opened 2020-08-22
Release/4.2.3

CompuCell3D:
This is a release PR. Most changes have already been reviewed. It is worth to do a quick scan to make sure things look reasonable


tjsego:
@CompuCell3D can we restore release/4.2.3? I think this would help with versioning.

PR #344, opened 2020-09-03
Made seeding of main application use higher clock resolution

tjsego:


jpsluka:
Maybe we should do the same for the output files. On a cluster it is easy to get more than one job with the same time based output folder name. Perhaps add milliseconds to the file/folder name.

CompuCell3D (approved):
looks good

CompuCell3D:
Merged the PR

Issue #350, opened 2020-09-06
fix all distance functions in NumericalUtils


CompuCell3D:
call some of them unconditional_invariant, and those that are not "unconditional" should respect boundary conditions

CompuCell3D:
Fixed - closing

Issue #347, opened 2020-09-05
fix distanceInvariantCM to obey periodic boundary conditions


CompuCell3D:

CompuCell3D:
Fixed - closing

Issue #312, opened 2020-08-10
FPP links rendering


CompuCell3D:
The links are NOT drawn properly if you chance the view to 3D or the plane of view (they are always rendered as if in x-y view no matter what)

CompuCell3D:
Many people noticed odd behavior with FPP in CC3D during the class. FPP edges weren’t being displayed properly in player, the edges seem to come and go randomly and other odd behaviors.

 

I have a simulation now that appears to be changing the membership in the cell list by type data. For example;

 

        for cell1 in self.cell_list_by_type(self.ACELL):

            print("\t\t\t cell1.id,",cell1.id)

            for cell2 in self.cell_list_by_type(self.BCELL):

                d=cell_cell_dist(cell1,cell2)

                self.focalPointPlasticityPlugin.createFocalPointPlasticityLink(cell1, cell2, 10, d, 500)

                print("\t\t\t    cell2.id",cell2.id)

 

# calculate the distance between two cells, cell1 and cell2

def cell_cell_dist(cell1,cell2):

    import math

    d=math.sqrt((cell1.xCOM-cell2.xCOM)**2+(cell1.yCOM-cell2.yCOM)**2+(cell1.zCOM-cell2.zCOM)**2)

    return(d)

 

Gives as output:

 

                Len acell and bcell lists: 4 5

                  acell lists:

1 2 5 8

                  bcell lists:

3 4 6 7 9

                         cell1.id, 1

                            cell2.id 3

                            cell2.id 4

                            cell2.id 6

                            cell2.id 7

                            cell2.id 9

                         cell1.id, 3

                            cell2.id 3

                            cell2.id 4

                            cell2.id 6

                            cell2.id 7

                            cell2.id 9

                         cell1.id, 3

                            cell2.id 3

                            cell2.id 4

                            cell2.id 6

                            cell2.id 7

                            cell2.id 9

                         cell1.id, 4

                            cell2.id 3

                            cell2.id 4

                            cell2.id 6

                            cell2.id 7

                            cell2.id 9

                         cell1.id, 9

                            cell2.id 3

                            cell2.id 4

                            cell2.id 6

                            cell2.id 7

                            cell2.id 9

 

                Len acell and bcell lists: 4 5

                  acell lists:

1 2 5 8

                  bcell lists:

3 4 6 7 9

 

[...]

PR #353, opened 2020-09-11
Feature/fpp links rendering fix

CompuCell3D:
Please checkout branch on your machine and run few simulations with FPP plugin to make sure the behavior is reasonable

This is PR addresses several issues

1) it improves rendering of fpp links
2) it changes how we compute distance in the fpp plugin to obey boundary conditions. Previous code was not correct in this regard. However, when the length of the link was much smaller than "avg lattice size" (as it was in most simulations) this did not create an issue. The issue occurred when e.g. you have link in the z direction and its length is > 0.5 of the z dimension.
3) Changes to invariant_distance/invariant_distance_vector to respect boundary conditions. Same motivation , problems, and impact as in 2)
 
Link rendering can still be improved but overall this version is much better and than what we had before

Issues addressed in this PR:
https://github.com/CompuCell3D/CompuCell3D/issues/350
https://github.com/CompuCell3D/CompuCell3D/issues/347
https://github.com/CompuCell3D/CompuCell3D/issues/312


CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityPlugin.cpp
```diff
@@ -250,76 +231,30 @@ void FocalPointPlasticityPlugin::update(CC3DXMLElement *_xmlData, bool _fullInit
...
 	}
 
 	plastParamsArray.clear();
 	plastParamsArray.assign(size,vector<FocalPointPlasticityTrackerData>(size,FocalPointPlasticityTrackerData()));
 
 	for(int i = 0 ; i < cellTypesVector.size() ; ++i)
 		for(int j = 0 ; j < cellTypesVector.size() ; ++j){
-			//cerr<<"cellTypesVector[i]="<<(int)cellTypesVector[i]<<endl;
-			//cerr<<"cellTypesVector[j]="<<(int)cellTypesVector[j]<<endl;
+			cerr<<"cellTypesVector[i]="<<(int)cellTypesVector[i]<<endl;
```
JulianoGianlupi, line 250: Do we want to keep the prints?

JulianoGianlupi (approved):

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityPlugin.cpp
CompuCell3D: good point. Fixed

CompuCell3D/core/CompuCell3D/Boundary/BoundaryStrategy.cpp
```diff
@@ -84,9 +86,17 @@ BoundaryStrategy::BoundaryStrategy(string boundary_x, string boundary_y,
...
 
+    boundaryConditionIndicator.assign(3, 0);
+
 	strategy_x = BoundaryFactory::createBoundary(boundary_x);
 	strategy_y = BoundaryFactory::createBoundary(boundary_y);
 	strategy_z = BoundaryFactory::createBoundary(boundary_z);
+
+    boundaryConditionIndicator[0] = boundary_x == "Periodic" ? 1 : 0;
+    boundaryConditionIndicator[1] = boundary_y == "Periodic" ? 1 : 0;
+    boundaryConditionIndicator[2] = boundary_z == "Periodic" ? 1 : 0;
```
tjsego, line 97: Would it be better to declare labels for all boundary condition types and assign them here instead?

tjsego (commented):
Just one comment to address. Otherwise this looks good.

CompuCell3D/core/CompuCell3D/Boundary/BoundaryStrategy.cpp
CompuCell3D: Agree, we can make a bigger PR to fix hardcoded strings

CompuCell3D: added new issue to address it https://github.com/CompuCell3D/CompuCell3D/issues/358

Issue #241, opened 2020-05-08
Report internal pressure


JulianoGianlupi:
Add automatic calculation of internal pressure,

``` p = 2 lambda * (volume - target_volume)```

Have it as cell property? Probably a good idea

tjsego:
Resolved by #354

PR #354, opened 2020-09-14
General class for derived properties

tjsego:
Highlights: 

- New general class: DerivedProperty
- DerivedProperty members return a value like a member variable according to a member function evaluation defined on the parent class in C++ and Python
- A member function defining the value of a DerivedProperty is restricted to information intrinsic to its class
- New interface file collects all Python implementations and includes a macro that reduces implementation to a single line per DerivedProperty
- Implementations in this PR: CellG pressure, surface tension and cluster surface tension

Notes on future work: waiting on link class definition to define a link tension

### DIFF
--- a/CompuCell3D/core/CompuCell3D/CMakeLists.txt
+++ b/CompuCell3D/core/CompuCell3D/CMakeLists.txt
@@ -29,6 +29,7 @@ SET(SRCS
   PluginManager.cpp
   Simulator.cpp
   ClassRegistry.cpp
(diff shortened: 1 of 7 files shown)


tjsego:
Resolves #241

CompuCell3D (approved):

CompuCell3D:
Can we add a demo showing how this new functionality works?. Otherwise looks good

JulianoGianlupi (approved):

Issue #348, opened 2020-09-05
Allow different links to have different constituent laws. 


CompuCell3D:
Pre-configure several standard formulas. 
Allow this to be configurable from Python including specifying custom link law from Python

maciekswat:
Fixed

Issue #345, opened 2020-09-05
Make sure that fppd attributes cannot be altered


CompuCell3D:
e.g.

fppd.targetDistance should not be allowed. Only allowed via API call that modifies two fppds that describe the link

tjsego:
Resolved by #361.

Issue #302, opened 2020-08-10
Refactor FPP plugin to include more flexible/convenient link definitions


CompuCell3D:
We need much clearer explanations of how to create links and what they do. We should be able to define multiple classes of links and have clear ways of defining their energy functionals. At a minimum, we should be able to define a link type and then attach a link of a given type between any two cells. Having multiple links between the same cells with different energies is very powerful. Setting the parameters for individual links is currently confusing and needs to be better documented. 

I specifically want to be able to automatically generate links of the following type--nearest neighbors in a chain with target lengths equal to the current length. Second neighbor links in a chain with energies equal to arccos of (l-ltarget) or arcos^2 of (l-ltarget) to give bending moduli

jpsluka:
We really need to describe the difference between links and FPP and ...

I would also add one more step beyond angles: dihedrals. (1-4 torsions instead of just 1-3 angles)

tjsego:
Should be resolved by #361. Can we close this?

tjsego:
Closing. We can open new issues that target specific aspects of the new interface.

Issue #83, opened 2017-08-30
Allow association of constitutive FPP links laws with particular link 


CompuCell3D:
At the moment the "link law" can only be specified for the entire FPP plugin we may want to associate different laws with particular links

maciekswat:
fixed. Closing

PR #361, opened 2020-10-05
Feature/fpp overhaul

tjsego:
### Overhaul of focal point plasticity

Summary

- New link classes now form the basis of link computations
- Three link classes are defined (regular, internal and anchor links) and derived from a base class with an interface for link-specific model specification.
- Link inventory classes function like the cell inventory, but for each type of link; each link inventory is derived from a base class through which most link management is defined. 
- New Python and wrapped C++ infrastructure simplifies link creation/destruction and provides convenience functions and iterators for links by inventory and cell.
- Maintains all current functionality of focal point plasticity. 
- Supports multi-threading to the same extend as the current implementation. 
- Performance in single- and multi-threaded simulations is approximately the same as the current implementation.


tjsego:
Concerning #345, link definition allows the user to safely specify properties per link.

tjsego:
Concerning #348, basic infrastructure is in place to specify constitutive laws per link in Python (FocalPointPlasticityLinkBase::setConstitutiveLaw). However, it hasn't yet been integrated into the plugin and tested.

tjsego:
Resolves much of #302.

tjsego:
> Concerning #348, basic infrastructure is in place to specify constitutive laws per link in Python (FocalPointPlasticityLinkBase::setConstitutiveLaw). However, it hasn't yet been integrated into the plugin and tested.

Same for #83.

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityPlugin.cpp
```diff
@@ -206,436 +184,451 @@ void FocalPointPlasticityPlugin::update(CC3DXMLElement *_xmlData, bool _fullInit
...
+	//initializing maxNumberOfJunctionsTotalVec based on plastParamsArray .maxNumberOfJunctionsTotalVec is indexed by cell type  	
+	maxNumberOfJunctionsTotalVec.assign(size, 0);
+	for (int idx = 0; idx<maxNumberOfJunctionsTotalVec.size(); ++idx) {
+		int mNJ = 0;
+		for (int j = 0; j < maxNumberOfJunctionsTotalVec.size(); ++j) {
+
+			mNJ += plastParamsArray[idx][j].maxNumberOfJunctions;
 		}
+		maxNumberOfJunctionsTotalVec[idx] = mNJ;
+		cerr << "maxNumberOfJunctions for type " << idx << " is " << maxNumberOfJunctionsTotalVec[idx] << endl;
```
JulianoGianlupi, line 244: Do you want to keep this print?

JulianoGianlupi (approved):

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityPlugin.cpp
tjsego: Nope. Good catch. Removed in 53f32d4.

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
```diff
@@ -0,0 +1,233 @@
...
+	class FPPLinkListBase : public std::vector<T*> {
+	public:
+		typedef typename std::vector<T*>::iterator FPPLinkListIterator_t;
+		virtual ~FPPLinkListBase() {}
+	};
+
+	class FOCALPOINTPLASTICITY_EXPORT FPPLinkID {
+	public:
+		long id0;
+		long id1;
```
CompuCell3D, line 55: if we go with std::unordered_map you will need to create hashing function too

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
```diff
@@ -0,0 +1,233 @@
...
+			}
+		}
+		// Remove a link without any additional internal work
+		void removeLinkNoChain(LinkType* _link) {
+			const FPPLinkID linkId = getLinkId(_link);
+			linkInventoryItr_t itr = linkInventory.find(linkId);
+			if (itr != linkInventory.end()) linkInventory.erase(linkId);
+		}
+		// Remove tracker data from cells
+		void removeTrackerData(LinkType* _link) {
```
CompuCell3D, line 139: Since we are overhauling everything I woudl get rid of TrackerData from cells. This means that no cell should have ANY reference to FPPTrackerData. Instead we should write a code that "simulates existence" of FPPTrackerData and I think you are doing it already. I want to make sure that we move away from this implicit link definition that we currently have

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
```diff
@@ -0,0 +1,233 @@
...
+				itr = fppdSet.find(fppd);
+				if (itr != fppdSet.end()) fppdSet.erase(fppd);
+			}
+		}
+
+		FPPInventory_t* getCellLinkInventory(CellG* _cell) { 
+			return &cellLinkInventoryTrackerAccessor->get(_cell->extraAttribPtr)->linkInv;
+		}
+
+		CellG* getCellGFromConst(const CellG* _cell) {
```
CompuCell3D, line 161: In general std::const_cast<CellG*> should be used - this si standard C++ way of dealing with constantness. I also noticed a lot of overloaded constructors in one of the class that differ only by the fact that some of the args are constants and some are not. I would avoid this typo of overloads because they make code hard to read

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
```diff
@@ -0,0 +1,233 @@
...
+
+		FPPLinkInventoryTracker() {};
+		~FPPLinkInventoryTracker() {};
+		
+		FPPLinkInventoryBase<LinkType> linkInv;
+	};
+
+
+	template <class LinkType>
+	class FOCALPOINTPLASTICITY_EXPORT FPPLinkInventoryBase {
```
CompuCell3D, line 89: If possible I would create just one Inventory class that handles all link types I really dont' think there is a need to have template and a base class for relatively simple concept like inventory of links. In a sense inventory of links is a container that holds objects of type link. And all links are of the same type (with anchor being tiny exception)

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
```diff
@@ -0,0 +1,233 @@
...
+
+		virtual int getLinkInventorySize() { return linkInventory.size(); }
+		linkInventory_t& getContainer() { return linkInventory; }
+		virtual linkInventoryItr_t linkInventoryBegin() { return linkInventory.begin(); }
+		virtual linkInventoryItr_t linkInventoryEnd() { return linkInventory.end(); }
+		virtual void incrementIterator(linkInventoryItr_t& _itr) { ++_itr; }
+		virtual void decrementIterator(linkInventoryItr_t& _itr) { --_itr; }
+
+		// Add a link to the link inventory and update internals
+		void addToInventory(LinkType* _link) {
```
CompuCell3D, line 192: I woudl imagine that this function woudl take a link , add it to inventory and this is it. Why are we calling those extra functions?

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
```diff
@@ -0,0 +1,233 @@
...
+		// Get link inventory list
+		FPPLinkList getLinkList() {
+			FPPLinkList fppLinkList;
+			for (linkInventoryItr_t itr = linkInventory.begin(); itr != linkInventory.end(); ++itr)
+				fppLinkList.push_back(itr->second);
+			return fppLinkList;
+		}
+		// Get link inventory list by cell
+		FPPLinkList getCellLinkList(CellG* _cell) { return getCellLinkInventory(_cell)->getLinkList(); }
+		FPPLinkList getCellLinkList(const CellG* _cell) {
```
CompuCell3D, line 216: Do not oveload functions based on constantness

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
CompuCell3D: I would remove this function completely. It does not belong to inventory

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinks.h
```diff
@@ -0,0 +1,300 @@
...
+			initiator = _initiator;
+			initiated = _initiated;
+			potts = _potts;
+			fppltd = FocalPointPlasticityLinkTrackerData(_fppltd);
+			fppltd.anchor = false;
+
+			DerivedProperty<FocalPointPlasticityLinkBase, float, &FocalPointPlasticityLinkBase::getDistance> length(this);
+			DerivedProperty<FocalPointPlasticityLinkBase, float, &FocalPointPlasticityLinkBase::getTension> tension(this);
+		}
+		FocalPointPlasticityInternalLink(const CellG *_initiator, CellG *_initiated, Potts3D *_potts, FocalPointPlasticityLinkTrackerData _fppltd) :
```
CompuCell3D, line 227: I would avoid overloads on constantness

CompuCell3D: There is way to many construtors here

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinks.h
```diff
@@ -0,0 +1,300 @@
...
+		FocalPointPlasticityAnchor(CellG *_cell, Potts3D *_potts, float _lambdaDistance = 0.0, float _targetDistance = 0.0, float _maxDistance = 100000.0, int _initMCS = 0, std::vector<float> _anchorPoint = std::vector<float>(3, 0.0)) :
+			FocalPointPlasticityAnchor(_cell, _potts, FocalPointPlasticityLinkTrackerData(_lambdaDistance, _targetDistance, _maxDistance, _initMCS)) {}
+		FocalPointPlasticityAnchor(const CellG *_cell, Potts3D *_potts, float _lambdaDistance = 0.0, float _targetDistance = 0.0, float _maxDistance = 100000.0, int _initMCS = 0, std::vector<float> _anchorPoint = std::vector<float>(3, 0.0)) :
+			FocalPointPlasticityAnchor(getCellGFromConst(_cell), _potts, _lambdaDistance, _targetDistance, _maxDistance, _initMCS, _anchorPoint) {}
+
+		const long getId0() { return initiator->id; }
+		const long getId1() { return fppltd.anchorId; }
+
+		// Get anchor point
+		std::vector<float> getAnchorPoint() { return fppltd.anchorPoint; }
```
CompuCell3D, line 290: So really, the only thing that warrans existence of those 3 classes of links is anchor point. But concenptually link is a linik "regular" link hooks to a moving target anchor link is hooked to stationary target. This is the only difference and I feel like we should have just one class for those 3 types of links External, interna, anchor. My motivation is code simplicity although I can see how you can make argument to keep 3 classes

CompuCell3D: I think we we add getAnchorPoint() setAnchorPoint getAnchorId to  "unified" links class and throw exception when user tries to set e.g. anchorPoint on a non-anchor link this woudl be a decent compromise and the code woudl be much simpler. What do you think?

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityPlugin.cpp
```diff
@@ -206,436 +184,445 @@ void FocalPointPlasticityPlugin::update(CC3DXMLElement *_xmlData, bool _fullInit
...
+
+	if (((int)plastParamsArray.size()) - 1<newCell->type) { //the newCell type is not listed by the user
+		newJunctionInitiatedFlag = false;
 		return 0.0;
 	}
 
-	if(focalPointPlasticityTrackerAccessor.get(newCell->extraAttribPtr)->focalPointPlasticityNeighbors.size()>=maxNumberOfJunctionsTotalVec[newCell->type]){
-		newJunctionInitiatedFlag=false;
+	//check if new cell can accept new junctions
+	if (linkInv.getNumberOfJunctions(const_cast<CellG*>(newCell)) >= maxNumberOfJunctionsTotalVec[newCell->type]) {
```
CompuCell3D, line 441: It looks like we need to keep max number of junctions as a cell attribute. Everythign from FPP tracker data can and sohuld  be moved to separate link object  but maxNUmber of junctions must stay within a cell

CompuCell3D: Or perhaps we can get away with it and it looks like what I mentioned is not necessary

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityPlugin.cpp
```diff
@@ -649,310 +636,241 @@ double FocalPointPlasticityPlugin::changeEnergy(const Point3D &pt,const CellG *n
 	Coordinates3D<float> centMassNewBefore;
 
 
-	if(oldCell){
-		centMassOldBefore.XRef()=oldCell->xCM/(float)oldCell->volume;
-		centMassOldBefore.YRef()=oldCell->yCM/(float)oldCell->volume;
-		centMassOldBefore.ZRef()=oldCell->zCM/(float)oldCell->volume;
+	if (oldCell) {
+		centMassOldBefore.XRef() = oldCell->xCM / (float)oldCell->volume;
```
CompuCell3D, line 640: It woudl be great to fix this ugly syntax centMassOldBefore.YRef() = ... we do not have to do it now but I dont like it even though I wrote it ;)

CompuCell3D/core/Demos/PluginDemos/FocalPointPlasticityLinks/Simulation/FocalPointPlasticityLinksSteppables.py
```diff
@@ -0,0 +1,93 @@
...
+            print(f'neighbor_order: {link1j.getNeighborOrder()}, {fppd.neighborOrder}')
+            print(f'init_mcs: {link1j.getInitMCS()}, {fppd.initMCS}')
+
+            # Let's check the link's derived properties
+            print(f'length: {link1j.length}')
+            print(f'tension: {link1j.tension}')
+
+        # Now let's quickly get the list of all cells linked to cell 1
+        linked_list = self.get_fpp_linked_cells(cell1)
+        [print(f'Cell 1 has a link with cell {c.id}') for c in linked_list]
```
CompuCell3D, line 51: why list comprehension with prints?

CompuCell3D/core/Demos/PluginDemos/FocalPointPlasticityLinks/Simulation/FocalPointPlasticityLinksSteppables.py
```diff
@@ -0,0 +1,93 @@
...
+        # Every 1000 steps, let's randomly select a type 1 - type 2 link and break it
+        if mcs % 1000 == 0:
+            if self.get_number_of_fpp_links() > 0:
+                link_list = [link for link in self.get_focal_point_plasticity_link_list()]
+                random.shuffle(link_list)
+                self.delete_fpp_link(link_list[0])
+
+        # Every 10000 steps, let's break all links
+        if mcs % 10000 == 0:
+            [self.remove_all_cell_fpp_links(cell, links=True) for cell in self.cell_list]
```
CompuCell3D, line 93: same here list comprehension without actully creating a list

CompuCell3D/core/Demos/PluginDemos/FocalPointPlasticityLinks/Simulation/FocalPointPlasticityLinksSteppables.py
```diff
@@ -0,0 +1,93 @@
...
+                celli, cellj = cell_list[0:2]
+                if self.get_number_of_fpp_junctions_by_type(celli, celli.type) == 0 and \
+                        self.get_number_of_fpp_junctions_by_type(cellj, cellj.type) == 0:
+                    new_link = self.new_fpp_link(celli, cellj, lambda_distance, target_distance, max_distance)
+                    new_link.setActivationEnergy(activation_energy)
+
+        # Every 1000 steps, let's randomly select a type 1 - type 2 link and break it
+        if mcs % 1000 == 0:
+            if self.get_number_of_fpp_links() > 0:
+                link_list = [link for link in self.get_focal_point_plasticity_link_list()]
```
CompuCell3D, line 87: will it return a list of all links? (anchors, internal, external)?

CompuCell3D/core/Demos/PluginDemos/FocalPointPlasticityLinks/Simulation/FocalPointPlasticityLinksSteppables.py
```diff
@@ -0,0 +1,93 @@
...
+import random
+
+
+class FocalPointPlasticityLinksSteppable(SteppableBasePy):
+
+    def __init__(self, frequency=1):
+
+        SteppableBasePy.__init__(self, frequency)
+
+    def step(self, mcs):
```
CompuCell3D, line 12: I think this code looks much nicer than any previos code that involved links

CompuCell3D (changes requested):
I like most of the changes but I woudl simplify code - specifically avoid templetization and creating unnecesary inheritance hierarchies even thuogh technically it might make sene to do so but I feel that for this particular case the overcomplication of the code (templates/inheritance) is just not worh it

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
```diff
@@ -0,0 +1,233 @@
...
+
+	template <class LinkType>
+	class FOCALPOINTPLASTICITY_EXPORT FPPLinkInventoryBase {
+
+	public:
+
+		typedef FPPLinkListBase<LinkType> FPPLinkList;
+		typedef FPPLinkInventoryBase<LinkType> FPPInventory_t;
+
+		typedef std::map<const FPPLinkID, LinkType*> linkInventory_t;
```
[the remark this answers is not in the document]
tjsego: Noted. This will make for a nice improvement.

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
```diff
@@ -0,0 +1,233 @@
...
+	/**
+	Written by T.J. Sego, Ph.D.
+	*/
+
+	class Potts3D;
+	class FocalPointPlasticityTracker;
+
+	typedef std::set<FocalPointPlasticityTrackerData> FPPTrackerDataSet;
+
+	template <typename T>
```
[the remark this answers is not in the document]
tjsego: No opinion on this one. Mostly just giving a name to a vector of various link types, like the typedefs for iterators and containers in (CellInventory.h)[https://github.com/CompuCell3D/CompuCell3D/blob/feature/fpp_overhaul/CompuCell3D/core/CompuCell3D/Potts3D/CellInventory.h], with some basic support for future dev if we so choose.

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
```diff
@@ -0,0 +1,233 @@
...
+		}
+		~FPPLinkID() {}
+		bool operator < (const FPPLinkID & _rhs) const { return id0 < _rhs.id0 || (id0 == _rhs.id0 && id1 < _rhs.id1); }
+		bool operator == (const FPPLinkID & _rhs) const { return id0 == _rhs.id0 && id1 == _rhs.id1; }
+		bool operator != (const FPPLinkID & _rhs) const { return !(operator==(_rhs)); }
+	};
+
+	template <class LinkType> class FPPLinkInventoryBase;
+
+	template <class LinkType>
```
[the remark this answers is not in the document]
tjsego: This I think these generic elements are important for a few reasons. 

First, while much of link functionality is the same for all link types, there are some particularities about the different link types (specifically external/internal vs. anchor) that change the way we interact with the data that's managed by the individual inventories. For example, we can ask an inventory that manages external or internal link what all cells are linked to a particular cell (according to the link type), which we can't ask of an anchor inventory. Likewise, we can assign the role of getting the next anchor id to the anchor inventory, but putting this function in the external or internal inventory definition is confusing. 

The same is true for the hierarchy of class definitions for the link types. Much is the same for all link types (e.g., FocalPointPlasticityLinkBase::getMaxDistance()), while some are specific to link type (e.g., FocalPointPlasticityAnchor::getAnchorPoint()). Again, we could just define a uniform interface for all link types and throw errors or something comparable when a function is used inappropriately, but this also seems to add more confusion than the complexity of this code, and I expect that it would also make inventory integration much more complicated in both the frontend and backend. 

Second, this structure should well support future dev for links. It was tempting to just assume that external and internal links will function the same, and then add some tricks to support anchors. But this assumes that the complexity of the definitions of the link types will not increase, and also that new link types will not be added in future work. The approach used in the PR is important to both of these points in that we can (hopefully) avoid a lot of supporting work for development of links by adopting a general approach now.

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
tjsego: Simulating the existence of FPPTrackerData is exactly what this is intended to do. Currently the reference to FPPTrackerData is still defined for each cell, though the data is now managed by the link inventories. To my knowledge, the only thing this now supports is the Python iterators for the data list. However, we can generate the data for those iterators from the link inventories. I'll make this change and remove FPPTrackerData from the cells altogether before we merge.

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
tjsego: See my commentary above about the approach adopted here. In general, I agree that simpler is better.

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
tjsego: addTrackerData manages the FPPTrackerData for each cell, and will be removed soon. The remaining three calls actually do the adding of a link to the container (addLinkNoChain). You'll notice that each cell has its own link inventory of the same link type (accessed through getCellLinkInventory). This makes for quick accessing of links by cell.

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinks.h
tjsego: See my commentary above. In general, I'm with you on simplicity. Handling the subtle differences between link types seemed better suited for their class definitions, rather than everywhere they are integrated. The generic architecture also seemed like a better approach for supporting future dev of link definitions and new link types. I'm also not particularly fond of putting members on class definitions that are only relevant for particular interpretations of the class instance (e.g., putting getAnchorPoint() on an external link type). It seems like this might cause more problems in the future for us and the user than the added value of simpler class code.

CompuCell3D/core/Demos/PluginDemos/FocalPointPlasticityLinks/Simulation/FocalPointPlasticityLinksSteppables.py
tjsego: Just demonstrating new functionality. No preference on how we show it off.

CompuCell3D/core/Demos/PluginDemos/FocalPointPlasticityLinks/Simulation/FocalPointPlasticityLinksSteppables.py
tjsego: Just a concise way of removing all links. We could instead demo 

```python
for cell in self.cell_list:
    self.remove_all_cell_fpp_links(cell, links=True)
```

CompuCell3D/core/Demos/PluginDemos/FocalPointPlasticityLinks/Simulation/FocalPointPlasticityLinksSteppables.py
tjsego: This one returns a list of all external links. `get_focal_point_plasticity_internal_link_list` does the same for internal links, and `get_focal_point_plasticity_anchor_list` for anchors. Their [lists and iterators](https://github.com/CompuCell3D/CompuCell3D/blob/feature/fpp_overhaul/cc3d/core/iterators.py) are derived from `_FocalPointPlasticityLinkListBase` and `_FocalPointPlasticityDataIteratorBase`. As discussed above, specialization is employed by link type to support fancy features in future work.

CompuCell3D/core/Demos/PluginDemos/FocalPointPlasticityLinks/Simulation/FocalPointPlasticityLinksSteppables.py
tjsego: Nice!

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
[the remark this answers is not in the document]
tjsego: Implemented in https://github.com/CompuCell3D/CompuCell3D/pull/361/commits/26d759b42d4cf99df801430ad6aeee84928218a4

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
tjsego: Implemented using Cantor pairing function in https://github.com/CompuCell3D/CompuCell3D/pull/361/commits/26d759b42d4cf99df801430ad6aeee84928218a4

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
tjsego: Implemented in https://github.com/CompuCell3D/CompuCell3D/pull/361/commits/4ce8f702eb17e2ca303319d02b23df9676e47ed4

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityPlugin.cpp
tjsego: Implemented in https://github.com/CompuCell3D/CompuCell3D/pull/361/commits/66ddddbcbb07ec9f6d707e591dda1ecf0d7aeae3

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
tjsego: Implemented in https://github.com/CompuCell3D/CompuCell3D/pull/361/commits/1305c536c83539572ee8f043066638c4d3c5e949, https://github.com/CompuCell3D/CompuCell3D/pull/361/commits/5e10487f5061e77cf90a34f83bd8d9f5231628a0, and https://github.com/CompuCell3D/CompuCell3D/pull/361/commits/9e3315490aa05a7c7cf8c66012cfe12c3ba4e0c6

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinkInventoryBase.h
tjsego: Implemented in https://github.com/CompuCell3D/CompuCell3D/pull/361/commits/5e10487f5061e77cf90a34f83bd8d9f5231628a0

CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinks.h
tjsego: Implemented in https://github.com/CompuCell3D/CompuCell3D/pull/361/commits/1305c536c83539572ee8f043066638c4d3c5e949

CompuCell3D (approved):
looks good, tests pass. Let's merge to develop

PR #375, opened 2020-11-30
Small function for proper use cound on nanohub

JulianoGianlupi:
Count\*

It's already part of an older release (we had to go fast), adding it to develop so we don't accidentally loose it

### DIFF
--- a/cc3d/CompuCellSetup/cluster_utils.py
+++ b/cc3d/CompuCellSetup/cluster_utils.py
@@ -1,0 +1,27 @@
+
+def check_nanohub_and_count():
+    """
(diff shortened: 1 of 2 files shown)


CompuCell3D (changes requested):
Looks good. The only suggestion I have is to move this function to a new module (new python script in the same folder as simulation setup). Let's call it cluster_utils.py or cluster_utilities.py - or whatever other name you think is appropriate. This way it will be easier to keep code organized

JulianoGianlupi:
Good point, implemented

CompuCell3D (approved):
Looks good

PR #376, opened 2020-12-21
Fixed improper access of flag for steering parameters

tjsego:
Minor bug fix.


cc3d/core/PySteppables.py
```diff
@@ -755,7 +755,10 @@ def steering_param_dirty(self, name=None):
         with pg.steering_panel_synchronizer:
 
             if name is not None:
-                return self.get_steering_param(name=name).dirty_flag
+                try:
```
CompuCell3D, line 758: maybe we can introduce set_dirty_flag(param_name) function and move the code there?

CompuCell3D (approved):
looks ok. minor suggestion

PR #383, opened 2021-01-05
Simple function to get roadrunner of sbml model

JulianoGianlupi:
Don't know if it's too late for 4.2.4, we can leave it for later

Also not sure if I should have made it part of the SBMLSolverHelper class


tjsego (commented):
See `SBMLSolverHelper.get_sbml_simulator`. I'd rather just make that a class method than add a function that does the same thing.

JulianoGianlupi:
Done

tjsego (approved):

PR #387, opened 2021-01-08
correct directory handling fix, by Steven Clark

JulianoGianlupi:
By Clark Steven

We figured out that NH wasn't counting things right because of directory shenanigans


tjsego (approved):

tjsego:
@CompuCell3D it would be great to get this into v4.2.4 and then update on nanohub.

PR #396, opened 2021-02-12
disabled checking for updates if in NH

JulianoGianlupi:


cc3d/player5/ViewManager/SimpleViewManager.py
```diff
@@ -453,6 +453,11 @@ def check_version(self, check_interval=-1, display_no_update_info=False):
         :return:None
         """
 
+        from os import environ
```
CompuCell3D, line 456: move import to the top of the file

CompuCell3D (changes requested):

cc3d/player5/ViewManager/SimpleViewManager.py
CompuCell3D: otherwise it looks good!

tjsego (approved):

cc3d/player5/ViewManager/SimpleViewManager.py
JulianoGianlupi: done

JulianoGianlupi:
@CompuCell3D @tjsego can we also push this change to v4.2.4? that way NH will get the pertinent update sooner

PR #397, opened 2021-02-14
Feature/graphics 3d fix

CompuCell3D:
This PR addresses issues with "missing cells in 3D" https://github.com/CompuCell3D/CompuCell3D/issues/393. This happened only when NeighborTracker was enabled and was caused by "overly eager" optimization. Now when user choose drawing cell borders in 3D cell shell optimization (drawing only outermost shell of cells) is enabled . we do not choose to display borders in 3D thenn all the cells are shown (except you cannot see borders). We might want to revisit it later and have separate setting for cell shell optimization but for now this fix addresses the issue

### DIFF
--- a/cc3d/CompuCellSetup/simulation_setup.py
+++ b/cc3d/CompuCellSetup/simulation_setup.py
@@ -580,6 +580,10 @@ def main_loop_player(sim, simthread=None, steppable_registry=None):
 
     cur_step = beginning_step
 
+    # initial drawing - we use mcs = -2 as a sentinel for mcs right after start function of steppables is completed
+    simthread.loopWork(-2)
+    simthread.loopWorkPostEvent(-2)
+
     while cur_step < sim.getNumSteps():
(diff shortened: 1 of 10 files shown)


cc3d/CompuCellSetup/simulation_setup.py
```diff
@@ -580,6 +580,10 @@ def main_loop_player(sim, simthread=None, steppable_registry=None):
 
     cur_step = beginning_step
 
+    # initial drawing - we use mcs = -2 as a sentinel for mcs right after start function of steppables is completed
+    simthread.loopWork(-2)
```
tjsego, line 584: Any way we could formalize this?

tjsego: Specifically, using a defined variable instead of hard-coding a value for the flag.

tjsego (commented):
Only minor thoughts from me. Otherwise I think this looks good.

JulianoGianlupi (approved):

PR #400, opened 2021-02-21
Added safe cell velocity calculations

tjsego:
Used this approach instead of naïve implementation based on cell properties, to handle periodic boundary conditions.


CompuCell3D (approved):
Looks good. I wonder how useful this instantaneous velocity will be

PR #401, opened 2021-02-21
Feature/multiaxes plots

CompuCell3D:
This is an improvement to Plotting API that allows control of individual y-axes in the plot and allows users to define multiple y axes

### DIFF
--- a/cc3d/player5/Plugins/ViewManagerPlugins/PlotWindowInterface.py
+++ b/cc3d/player5/Plugins/ViewManagerPlugins/PlotWindowInterface.py
@@ -324,9 +324,6 @@ class PlotWindowInterface(QtCore.QObject):
                 y_axis_item.setLabel(plot_name, color=color)
 
                 y_axis_item.setLogMode(y_log_flag)
-                # if y_log_flag:
-                #     print('this should neve happen y_log_flag=', y_log_flag)
-                #     y_axis_item.setLogMode(y_log_flag)
 
             if set_y_range_allowed:
                 # print('this should never happen set_y_range_allowed=', set_y_range_allowed)
@@ -387,7 +384,6 @@ class PlotWindowInterface(QtCore.QObject):
 
             else:
                 self.pW.addItem(plot_obj)
-                # self.pW.plotItem.setYRange(0,20.)
 
             self.update_views()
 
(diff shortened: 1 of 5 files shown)


cc3d/player5/Plugins/ViewManagerPlugins/PlotWindowInterface.py
```diff
@@ -169,13 +154,44 @@ def set_data_bar_graph_item(self, plot_obj, x, y):
...
+        :param separate_y_axis:
+        :param y_min:
+        :param y_max:
+        :param y_scale_type:
+        :return:
+        """
+
+        if not separate_y_axis:
+            if y_min is not None or y_max is not None:
+                raise RuntimeError('y_min or y_max can only be set if you set separate_y_axis=True '
```
JulianoGianlupi, line 175: Is there any technical reason for this? If not I vote for letting the user set limits

CompuCell3D: you cannot be setting different limits for different data sets if you are using only single y_axes. It would make no sense to have first data series have limits 0..1 and second data series (on the same plot with shared y_axis) have limit from 2..3. Because this should would you choose then?

JulianoGianlupi: I may have misunderstood the function then. 

Right now the user can set fixed axis range in player by right clicking the plot and setting the y/x range, my assumption was that this (without the raise error) could be used to do the same from python (I can't recall if that's already possible)

CompuCell3D: True (about right-click) but this PR lets you do a bit more programmatically and having multiple axes was not available before.

JulianoGianlupi: >  lets you do a bit more programmatically

That was what I was getting at, (if I understood the code correctly) this PR wouldn't allow the user to set the y limits from python if they are doing a single axis but would allow it for multiple.

CompuCell3D: Exactly, I think this is good. Let me know if you agree or if we should change the behavior (we can always change the behavior later as well)

tjsego (approved):

PR #412, opened 2021-03-12
added self.merge_cells to cell manipulation tweedit menu

JulianoGianlupi:
Added self.merge_cells to cell manipulation tweedit menu


CompuCell3D (approved):
looks OK, merging

PR #418, opened 2021-03-23
fix/CellTypeColorMapModel

tjsego:
Fix to occasional unhandled exception. 

`CellTypeColorMapModel.item_data` is an `OrderedDict`, so the exception to handle in `CellTypeColorMapModel.headerData` is a `KeyError`.


CompuCell3D (approved):
Good catch!

PR #419, opened 2021-03-25
Feature/two-sided chemotaxis

tjsego:
This feature specifically targets the mechanism of combined chemotactic migration, in that both cells at an intercellular interface can be chemotactically migrating and thus collectively contribute to motion of the interface.

### DIFF
--- a/CompuCell3D/core/CompuCell3D/plugins/Chemotaxis/ChemotaxisPlugin.cpp
+++ b/CompuCell3D/core/CompuCell3D/plugins/Chemotaxis/ChemotaxisPlugin.cpp
@@ -274,6 +274,8 @@ void ChemotaxisPlugin::update(CC3DXMLElement *_xmlData, bool _fullInitFlag){
 			algorithmPtr=&ChemotaxisPlugin::merksChemotaxis;
 		} else if(chemotaxisAlgorithm=="regular") {
 			algorithmPtr=&ChemotaxisPlugin::regularChemotaxis;
+		} else if (chemotaxisAlgorithm=="reciprocated") {
+			algorithmPtr=&ChemotaxisPlugin::reciprocatedChemotaxis;
 		}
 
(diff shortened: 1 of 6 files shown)


JulianoGianlupi (approved):

CompuCell3D/core/CompuCell3D/plugins/Chemotaxis/ChemotaxisPlugin.cpp
```diff
@@ -556,6 +558,82 @@ double ChemotaxisPlugin::regularChemotaxis(const Point3D &pt, const CellG *newCe
...
+			if(chemotaxisDataPtr && chemotaxisDataPtr->okToChemotact(newCell, oldCell)){ 
+				ChemotaxisPlugin::chemotaxisEnergyFormulaFcnPtr_t formulaCurrentPtr = 0;
+				formulaCurrentPtr = chemotaxisDataPtr->formulaPtr;
+				if(formulaCurrentPtr){
+					energy += (this->*formulaCurrentPtr)(fieldVec[i]->get(potts->getFlipNeighbor()), fieldVec[i]->get(pt), *chemotaxisDataPtr);
+					chemotaxisDone = true;
+				}
+			}
+
+			if(!chemotaxisDone && (int)oldCell->type < vecVecChemotaxisData[i].size()){
```
CompuCell3D, line 621: I would try to encapsulate lines 621-628 and 590-597 as a separate function and call it instead of repeating the code. In general I would introduce more functions like that. And as you can see that would apply to many CC3D plugins which means we do need need to do it now but slowly work towards eliminating code duplication from many CC3D modules

CompuCell3D (approved):
Looks good, a minor comment that applies not only here but to a large chunk of CC3D code. No need to take immediate action just do those fixes when you have some spare time

PR #421, opened 2021-03-26
Feature/link dict

tjsego:
This PR extends useful CellG features to link classes, including

- A general python dictionary attached to all link instances
- All basic SBML support extended to ODE models attached to links

### DIFF
--- a/CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinks.h
+++ b/CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinks.h
@@ -23,6 +23,11 @@
 #ifndef FOCALPOINTPLASTICITYLINKS_H
 #define FOCALPOINTPLASTICITYLINKS_H
 
+#ifndef PyObject_HEAD
+struct _object;
+typedef _object PyObject;
+#endif
+
 #include <CompuCell3D/CC3D.h>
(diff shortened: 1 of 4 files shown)


CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinks.h
```diff
@@ -174,6 +181,17 @@ namespace CompuCell3D {
 		CellG* getObj0() { return initiator; }
 		// Get second object
 		CellG* getObj1() { return initiated; }
+
+		// Python support
+
+		PyObject *pyAttrib;
+
+		PyObject *getPyAttrib() {
+#ifdef SWIGPYTHON
```
CompuCell3D, line 190: you might need to be careful here when it to memory usage. I dont remember exactly the issue but I think I was able to create memory leaks. I think there might be some explanation of this in the comments of PyAttributeAdder

CompuCell3D (changes requested):
I think it makes sense to attach dict to "primitive objects" (cells , links). I was wondering if it would make more sense to same way of adding dict to CellG and to Link objects. When we create CellG object we use factory method (and by the same token we would have object destructing methods) and we could potentially use same approach here as well and we would be able to share the code. Out of curiosity, what would be the use case for having dictionary attached to links?

CompuCell3D (approved):
let's do next PR to create base class with python dict being a core data mamber

PR #422, opened 2021-03-27
Feature/maboss

tjsego:
This PR integrates basic MaBoSS functionality for simulating Boolean models, and emulates RoadRunner/SBML functionality in CC3D python. Pending docs on all included functionality can be found [here](https://github.com/CompuCell3D/PythonScriptingManual/blob/update/maboss/docs/maboss.rst). 

Notes
-------

Requires update to Windows dependencies. A working conda environment for Windows is the same as for Linux, with the following additions, 
- pywin32 (pip)
- m2w64-dlfcn (msys2)

The MaBoSS libraries can be built on Windows with mingw64 and then targeted for building the rest of the CC3D package with MSVC. 

Current Issues
----------------
- Install needs tested on Linux and Mac
- Build scripts could probably be cleaned up
- Should we distribute pre-built binaries for building from source on Windows with MSVC?
- Update launch scripts and install package as necessary for conda environment on Windows

Future Work
--------------
- Current public release of MaBoSS requires rebuilding libraries for changing maximum Network sizes (fixed at 64 nodes in current implementation). Dynamic Network sizes is an experimental feature that should be incorporated once released by MaBoSS team
- Implement multithreading for loops over networks attached to cells in CellInventory
- Consider updating RoadRunner and Antimony integrations to a similar architecture for supporting them in C++.


CompuCell3D (changes requested):
Adding boolean networks is great. We needed them for a long time and I know other users have been implementing them by hand. I admit I have not gone through entire code but I have few questions:
1. is there an option to compile CC3D without MaBoSS? Let's say the user faces compilation issues on their operating system but they do not need boolean network. Can we somehow disable compilation? BTW this is the only reason I am asking this question
2. Would it be possible to use  similar approach as we did with roadrunner/ antimony integration? I.e. we provide "hooks" for integrating MaBoSS but pyMaBoSS is compiled separately from CC3D? What I am hinting here is possibility to include in CC3D only code that is written by us and have all 3rd party additions be compiled separately and integrated "at the Python level"

tjsego:
> Adding boolean networks is great. We needed them for a long time and I know other users have been implementing them by hand. I admit I have not gone through entire code but I have few questions:
> 
> 1. is there an option to compile CC3D without MaBoSS? Let's say the user faces compilation issues on their operating system but they do not need boolean network. Can we somehow disable compilation? BTW this is the only reason I am asking this question
> 2. Would it be possible to use  similar approach as we did with roadrunner/ antimony integration? I.e. we provide "hooks" for integrating MaBoSS but pyMaBoSS is compiled separately from CC3D? What I am hinting here is possibility to include in CC3D only code that is written by us and have all 3rd party additions be compiled separately and integrated "at the Python level"

I've revised the description of the PR according to recent development. The extensions PR allows for selective installation of the MaBoSS extension, and the built-in support implemented in this PR handles when the extension is missing.  

@CompuCell3D I completely agree with you about migrating all 3rd party code out, and only include built-in supporting features here. I'm planning to migrate the RoadRunner and Antimony integrations to the extensions repo as simple metapackages, for now, and we can think about a tighter integration when opportune. I would also like to do similar restructuring for Player and Twedit.

.gitignore
```diff
@@ -3,6 +3,12 @@ BionetSolver/0.0.6/include/sbmlsolver/config.h
 CompuCell3D/core/BasicUtils/BasicPluginManager.h
 CompuCell3D/core/pythonSetupScripts/Version.py
 RoadRunner/ThirdParty/libsbml/src/sbml/common/libsbml-config-common.h
+CompuCell3D/core/MaBoSS/engine/src/BooleanGrammar.cc
```
CompuCell3D, line 6: Do we need this? I thought that BionetSolver was not a part of CC3D anymore . But maybe you are repurposing it for something else

CompuCell3D (approved):
I think it looks good. i like all the changes and I like the lightweight integration. I assume that CC3D still will work fine even if Maboss module is not available (just like it works fine when libroadrunner is not available ). From the code it looks like this is indeed the case but wanted to double-check.  We will also need to upgrade dependencies for CC3D from Python 3.6 to python 3.7

.gitignore
tjsego: The MaBoSS-related entries where for generated scripts that are site-specific. I removed those and outdated entries for RoadRunner and BionetSolver in https://github.com/CompuCell3D/CompuCell3D/pull/422/commits/9fa97875db69d524585f4b0df9a2da543683d3cf and https://github.com/CompuCell3D/CompuCell3D/pull/422/commits/974cd0d5c08ca357c87e9926b4d300a843bc6b0d.

PR #423, opened 2021-03-27
Fix/memory leaks

tjsego:
Fixes memory leaks associated with creating and destroying subwindows in Player.

### DIFF
--- a/cc3d/core/GraphicsOffScreen/MVCDrawModel2D.py
+++ b/cc3d/core/GraphicsOffScreen/MVCDrawModel2D.py
@@ -18,8 +18,8 @@ MODULENAME = '----- MVCDrawModel2D.py:  '
 
 
 class MVCDrawModel2D(MVCDrawModelBase):
-    def __init__(self, boundary_strategy=None):
-        MVCDrawModelBase.__init__(self,boundary_strategy)
+    def __init__(self, boundary_strategy=None, ren=None):
+        MVCDrawModelBase.__init__(self, boundary_strategy=boundary_strategy, ren=ren)
 
         self.cellsMapper = None
         self.hex_cells_mapper = None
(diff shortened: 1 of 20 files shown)


cc3d/core/GraphicsOffScreen/MVCDrawModel2D.py
```diff
@@ -223,7 +223,7 @@ def init_axes_actors(self, actor_specs, drawing_params=None):
         axes_actor.SetYTitle(vertical_label)
         # axesActor.SetFlyModeToOuterEdges()
 
-        label_prop = axes_actor.GetLabelTextProperty(0)
+        # label_prop = axes_actor.GetLabelTextProperty(0)
```
CompuCell3D, line 226: I think we can safely remove commented code .

cc3d/core/GraphicsOffScreen/MVCDrawModelBase.py
```diff
@@ -29,6 +29,30 @@ def __init__(self, boundary_strategy):
...
+    @property
+    def ren(self):
+        try:
+            o = self._ren()
+        except TypeError:
+            o = self._ren
+        return o
+
+    @ren.setter
+    def ren(self, _i):
```
CompuCell3D, line 51: what is '_i' here? An integer? Let's add docstring to make things clear

cc3d/core/GraphicsOffScreen/MVCDrawViewBase.py
```diff
@@ -37,12 +37,35 @@ def __init__(self, _drawModel):
...
+    @property
+    def drawModel(self):
+        return self._drawModel()
+
+    @drawModel.setter
+    def drawModel(self, _i):
+        self._drawModel = ref(_i)
+
+    @property
+    def ren(self):
```
CompuCell3D, line 56: do we need ren method in both Model and View classes? Perhaps we do but if we could keep it in only one of the classes it would be better I think . Not sure what the right answer here is so just posing a question

cc3d/player5/Graphics/GraphicsFrameWidget.py
```diff
@@ -102,7 +104,15 @@ def __init__(self, parent=None, originatingWidget=None):
         self.renWin = self.qvtkWidget.GetRenderWindow()
         self.renWin.AddRenderer(self.gd.get_renderer())
 
-        self.metadata_fetcher_dict = {
+    @property
+    def metadata_fetcher_dict(self) -> typing.Dict[str, typing.Callable[[str, str], dict]]:
```
CompuCell3D, line 108: Since this si a property maybe we can rename it to `metadata_dict` - if it does not make sense ignore the comment

cc3d/player5/Graphics/PlotFrameWidget.py
```diff
@@ -58,6 +59,24 @@ def __init__(self, parent=None, **kwds):
...
+        Parent if any, otherwise None
+        """
+        try:
+            o = self._parentWidget()
+        except TypeError:
+            o = self._parentWidget
+        return o
+
+    @parentWidget.setter
+    def parentWidget(self, _i):
```
CompuCell3D, line 74: I like it!

CompuCell3D (changes requested):
It looks great. Some minor comments to look over not sure if all of them make sense but overall looks like it is ready to be merged

CompuCell3D (approved):

PR #437, opened 2021-05-06
Disabled distribution of optimization

tjsego:
The optimization functionality is clearly not currently supported. Let's stop distributing it until it works.


CompuCell3D (approved):
I think for this one let's leave the commented out section - I mean leave it as a comment . Does it mean that optimization is not working correctly?

tjsego:
> I think for this one let's leave the commented out section - I mean leave it as a comment . Does it mean that optimization is not working correctly?

Not on Windows, at least. It looks for python27 dependencies.

PR #440, opened 2021-05-11
Feature/parameter scan gui

CompuCell3D:
This PR implements parameter scan configuration GUI. It is actually much simpler GUI-based configuration process then before - you select variable, number etc that you want to scan (highlight it in twedit++) and then right-click and choose an option to add it to parameter scan and then the GUI will show up. The same process can be used to remove a given parameter from being scanned. Pretty simple workflow overall


tjsego (approved):
Looks great. Nice cleanup, too.

JulianoGianlupi (approved):

PR #442, opened 2021-05-21
Bug fix/make move cell fcn safe

### LINKED ISSUE #441: move_cell doesn't work correctly
move_cell doesn't work correctly. If the cell is displaced by less than the cell's width than any pixels that were common to both cell locations are set to medium.  Looks like the code is moving the cell to the new location then deleting all pixels from the original location (or setting them to medium) evn if those pixels are part of the cell in its new location.

Also, move_cell requires the PixelTracker plugin but that is not in the manual or in the code snippet that Twedit++ inserts.

Found on WIndows 10, CC3D 4.2.4 20210123

JulianoGianlupi:
fixes #441

### DIFF
--- a/cc3d/core/PySteppables.py
+++ b/cc3d/core/PySteppables.py
@@ -1898,6 +1898,9 @@ class SteppableBasePy(SteppablePy, SBMLSolverHelper, MaBoSSHelper):
         :param shift_vector: {tuple,  list, array, or CompuCell.Point3D} 3-element list specifying shoft vector
         :return: None
         """
+		
+		if not cell:
+			raise TypeError(f'Cannot move non existing cell. Expected cell to be CompuCell.CellG, got {type(cell)}')
 
         # we have to make two list of pixels :
         # used to hold pixels to delete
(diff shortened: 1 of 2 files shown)


cc3d/core/PySteppables.py
```diff
@@ -1935,7 +1940,10 @@ def move_cell(self, cell, shift_vector):
         # Now we will delete old pixels
         medium_cell = CompuCell.getMediumCell()
         for pixel in pixels_to_delete:
-            self.cell_field[pixel.x, pixel.y, pixel.z] = medium_cell
+			# Safe deletion. Don't delete the old pixel
```
tjsego, line 1943: Indents need reduced.

cc3d/core/PySteppables.py
```diff
@@ -1898,6 +1898,9 @@ def move_cell(self, cell, shift_vector):
         :param shift_vector: {tuple,  list, array, or CompuCell.Point3D} 3-element list specifying shoft vector
         :return: None
         """
+		
+		if not cell:
```
tjsego, line 1902: Indents need reduced.

cc3d/core/PySteppables.py
```diff
@@ -1915,6 +1918,8 @@ def move_cell(self, cell, shift_vector):
         # If we try to reassign pixels in the loop where we iterate over pixel data
         # we will corrupt the container so in the loop below all we will do is to populate the two list mentioned above
         pixel_list = self.get_cell_pixel_list(cell)
+		if pixel_list is None:
```
tjsego, line 1921: Indents need reduced.

tjsego (changes requested):

cc3d/core/PySteppables.py
CompuCell3D: Thanks @JulianoGianlupi  for looking into this. Let me see how  this behaves in the example that Jim mentioned.

CompuCell3D (approved):

PR #445, opened 2021-05-22
Another round of fixes for move_cell function - replaced linear list …

CompuCell3D:
…lookups with much faster set lookup and fixed comparison error when dealing with CompUCell3D.Point3D and a list of pixels. Code that uses python sets should be much faster

Addresses https://github.com/CompuCell3D/CompuCell3D/issues/441

### DIFF
--- a/cc3d/core/PySteppables.py
+++ b/cc3d/core/PySteppables.py
@@ -1891,6 +1891,14 @@ class SteppableBasePy(SteppablePy, SBMLSolverHelper, MaBoSSHelper):
     def moveCell(self, cell, shiftVector):
         return self.move_cell(cell=cell, shift_vector=shiftVector)
 
+    def point3d_to_tuple(self, pt: CompuCell.Point3D) -> tuple:
+        """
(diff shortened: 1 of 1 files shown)


cc3d/core/PySteppables.py
```diff
@@ -1891,6 +1891,14 @@ def get_cell_boundary_pixel_list(self, cell, neighbor_order=-1):
     def moveCell(self, cell, shiftVector):
         return self.move_cell(cell=cell, shift_vector=shiftVector)
 
+    def point3d_to_tuple(self, pt: CompuCell.Point3D) -> tuple:
```
tjsego, line 1894: Could we add this instead to the python definition for Point3D, Dim3D, etc.? Something like "to_tuple()". I feel like that would be much more generally useful and also save on a little additional clutter for this class definition.

tjsego (approved):
Minor comment here. Otherwise, looks good.

JulianoGianlupi (approved):

PR #447, opened 2021-05-23
Feature/cast methods

### LINKED ISSUE #446: add to_tuple to the python definition for Point3D, Dim3D instead of having it be standalone function

tjsego:
Adds simple method to python interface for Point3D to return coordinates as a tuple. Closes #446.

### DIFF
--- a/CompuCell3D/core/CompuCell3D/Field3D/Dim3D.cpp
+++ b/CompuCell3D/core/CompuCell3D/Field3D/Dim3D.cpp
@@ -22,3 +22,53 @@
 
 #include "Dim3D.h"
 using namespace CompuCell3D;
+
+Dim3D::Dim3D() : Point3D() {}
+
+Dim3D::Dim3D(const short x, const short y, const short z) : Point3D(x, y, z) {}
+
+Dim3D::Dim3D(const Dim3D &dim) : Point3D(dim) {}  
+
+Dim3D &Dim3D::operator=(const Dim3D pt) {
+    x = pt.x;
+    y = pt.y;
+    z = pt.z;
+    return *this;
+}
+
+Dim3D &Dim3D::operator+=(const Dim3D pt) {
+    x += pt.x;
+    y += pt.y;
+    z += pt.z;
+    return *this;
+}
+
+Dim3D &Dim3D::operator-=(const Dim3D pt) {
+    x -= pt.x;
+    y -= pt.y;
+    z -= pt.z;
+    return *this;
+}
+    
+bool Dim3D::operator==(const Dim3D pt) const {
+    return (x == pt.x && y == pt.y && z == pt.z);
+}
+
+bool Dim3D::operator!=(const Dim3D pt) const {
+    
+    return !(*this==pt);
+}
+
+bool Dim3D::operator<(const Dim3D  _rhs) const{
+    return x < _rhs.x || (!(_rhs.x < x)&& y < _rhs.y)
+        ||(!(_rhs.x < x)&& !(_rhs.y <y )&& z < _rhs.z);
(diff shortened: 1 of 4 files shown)


CompuCell3D/core/CompuCell3D/Field3D/Dim3D.cpp
```diff
@@ -22,3 +22,53 @@
...
+Dim3D::Dim3D(const Dim3D &dim) : Point3D(dim) {}  
+
+Dim3D &Dim3D::operator=(const Dim3D pt) {
+    x = pt.x;
+    y = pt.y;
+    z = pt.z;
+    return *this;
+}
+
+Dim3D &Dim3D::operator+=(const Dim3D pt) {
```
CompuCell3D, line 39: I guess all those functions were in the header file because we were hoping that  compiler might inline them. Perhaps new compiler would optimize it to the point that there is not difference between current and previous codes. Since you have those two branches Could you do a quick benchamrking to see if defining some functions in the header file even matters?

CompuCell3D/core/CompuCell3D/Field3D/Point3D.h
```diff
@@ -44,66 +42,47 @@ namespace CompuCell3D {
...
-    //virtual void readXML(XMLPullParser &in);
-    //virtual void writeXML(XMLSerializer &out);
-    //// End XMLSerializable interface
+    bool operator<(const Point3D _rhs) const;
     
     friend std::ostream &operator<<(std::ostream &stream, const Point3D &pt);
+
+    // Python support
+
+#ifdef SWIGPYTHON
```
CompuCell3D, line 81: I wonder if perhaps a better approach would be to define this function in the SWIG CompuCell.i interface file - I guess you would use %{extend } directive, something like that is used for CellG class. If this is not too much trouble let's try this approach. I think what I am driving at is that I would like to keep CC3D API separate from depending on Python too much. (yes I know we already have few references to Python in e.g. CellG but I think since we are using SWIG we may want to take advantage of the extend functionality that it offers). What do you think?

CompuCell3D (changes requested):
Overall it looks very good I had some minor comments that I think could help use better understand performance impacts (if any) / give use a slightly cleaner code

CompuCell3D/core/CompuCell3D/Field3D/Dim3D.cpp
tjsego: Using regression tests, inline functions here on average are 28.5% faster. From a standpoint of readability, I'll just migrate the functions back to the header file and implement them outside the class definitions.

CompuCell3D/core/CompuCell3D/Field3D/Point3D.h
tjsego: I prefer class definitions to be localized as much as possible, rather than maintaining a giant monolithic script like CompuCell.i. This could be implemented in a number of ways, however, whether like adding python dictionaries with preprocessor directives as in CellG.h, [FocalPointPlasticity.h](https://github.com/CompuCell3D/CompuCell3D/blob/develop/CompuCell3D/core/CompuCell3D/plugins/FocalPointPlasticity/FocalPointPlasticityLinks.h) and done here, or expanding our usage of swig files as in [DerivedProperty.i](https://github.com/CompuCell3D/CompuCell3D/blob/master/CompuCell3D/core/pyinterface/CompuCellPython/DerivedProperty.i). For now, I can just move the implementation to CompuCell.i, as suggested.

tjsego:
I'm closing this PR now and will issue a fresh one with the suggested revisions.

PR #448, opened 2021-05-24
Add to_tuple to Dim3D and Point3D, with minor cleanup

### LINKED ISSUE #446: add to_tuple to the python definition for Point3D, Dim3D instead of having it be standalone function

tjsego:
Adds simple method to python interface for Point3D and Dim3D to return components as a tuple. Reworked from #447. Closes #446.

### DIFF
--- a/CompuCell3D/core/CompuCell3D/Field3D/Dim3D.h
+++ b/CompuCell3D/core/CompuCell3D/Field3D/Dim3D.h
@@ -25,8 +25,6 @@
 
 #include "Point3D.h"
 
-//#include <XMLCereal/XMLSerializable.h>
-
 #include <string>
 
 namespace CompuCell3D {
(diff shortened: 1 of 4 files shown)


CompuCell3D/core/pyinterface/CompuCellPython/CompuCell.i
```diff
@@ -293,6 +296,13 @@ using namespace CompuCell3D;
     s<<(*self);
     return s.str();
   }
+
+%pythoncode %{
```
CompuCell3D, line 300: I think we could move this to SWIG interface - in CompuCell.i. . This  way we will keep C++ "pure" and do all pythonnic extensions like that in SWIG. I think Current swig file has examples of how to to it. I think CellG class is using %extend directive

CompuCell3D/core/CompuCell3D/Field3D/Dim3D.h
```diff
@@ -44,65 +42,72 @@ namespace CompuCell3D {
     /// Copy constructor
     Dim3D(const Dim3D &dim) : Point3D(dim) {}  
 
-    Dim3D &operator=(const Dim3D pt) {
-      x = pt.x;
-      y = pt.y;
-      z = pt.z;
-      return *this;
-    }
+    Dim3D &operator=(const Dim3D pt);
```
CompuCell3D, line 45: I would be curious to see if there is any performance degradation when we move function definitions from header to implementation file possibly sacrificing inlining. I suspect that the effect should be minor or none, but if you have a code handy that can do both styles of implementations maybe we can do quick benchmark?

CompuCell3D (changes requested):
Overall it looks very good. There are two minor tasks - benchmarking possible loss of inlining in Point3D and Dim3D and using %{extend in SWIG to implement to_tuple function inside CompuCell.i SWIG interface

CompuCell3D/core/pyinterface/CompuCellPython/CompuCell.i
tjsego: Maybe I'm not understanding this feedback, but is this requesting what's already implemented here? This addition is inside an %extend{} for Dim3D, and in CompuCell.i. Am I misunderstanding something?

CompuCell3D/core/CompuCell3D/Field3D/Dim3D.h
tjsego: See #447. I measured a pretty significant performance degradation using the regression test suite. But these functions inline in this PR, and the implementations are just below the class definitions to improve readability. So no change now in performance.

JulianoGianlupi (approved):

CompuCell3D/core/CompuCell3D/Field3D/Dim3D.h
CompuCell3D: make sense, I got confused by how github presents changes I thought those functions were in the *.cpp file

CompuCell3D/core/pyinterface/CompuCellPython/CompuCell.i
CompuCell3D: Right, I got fooled by how github presents changes. For some reason I thought this code was part of the *.cpp file

CompuCell3D (approved):
Looks good!

PR #450, opened 2021-05-25
Removed stop_simulation in some demos

### LINKED ISSUE #366: Remove self.stop_simulation() from all demos except demo of self.stop_simulation()
Three demos use code like:
        if mcs == 300:
            self.stop_simulation()
This should be removed. Only the one demo of the stop_simulation() function should actually use it. If the other demos need to only do a small number of MCS that should be defined in the normal way. I just had a user spend a couple hours trying to figure out why the maximum number of MCS in the Potts block was being ignored in a demo.

These two demos use stop_simulation():
bacterium_macrophage_new_style\Simulation\bacterium_macrophage_2D_steppables.py      
cellsort_project_py_step_new_style\Simulation\cellsort_2D_steppables.py

tjsego:
Closes #366.

### DIFF
--- a/CompuCell3D/core/Demos/bacterium_macrophage_new_style/Simulation/bacterium_macrophage_2D_steppables.py
+++ b/CompuCell3D/core/Demos/bacterium_macrophage_new_style/Simulation/bacterium_macrophage_2D_steppables.py
@@ -1,10 +1,7 @@
 from cc3d.core.PySteppables import *
 from cc3d import CompuCellSetup
-import numpy as np
-from scipy import special, optimize
 from scipy import integrate
 from numpy import exp
-from scipy import stats
 
 
 class InventoryCheckSteppable(SteppableBasePy):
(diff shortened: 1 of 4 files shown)


JulianoGianlupi (approved):

CompuCell3D/core/Demos/bacterium_macrophage_new_style/Simulation/bacterium_macrophage_2D_steppables.py
```diff
@@ -61,6 +57,3 @@ def step(self, mcs):
                     pass
 
         self.plot_win.show_all_plots()
-
```
CompuCell3D, line 64: I would leave it here. I added it on purpose here to illustrate how to stop simulation. What do you think @tjsego ?

CompuCell3D (changes requested):
I woudl leave the stop function. Curious what you think @tjsego

tjsego:
> I woudl leave the stop function. Curious what you think @tjsego

If the code is as intended, then that's fine with me. In this case, let's close this PR and close #366 accordingly.

PR #451, opened 2021-05-27
Fix/momentofinertia

tjsego:
- Fixes a bug in calculation of eccentricity. 
- Adds abstraction to MomentOfInertiaPlugin.

### DIFF
--- a/CompuCell3D/core/CompuCell3D/plugins/MomentOfInertia/MomentOfInertiaPlugin.cpp
+++ b/CompuCell3D/core/CompuCell3D/plugins/MomentOfInertia/MomentOfInertiaPlugin.cpp
@@ -175,28 +175,28 @@ void MomentOfInertiaPlugin::cellOrientation_xy(const Point3D &pt, CellG *newCell
 		lMinNew=0.5*(newCell->iXX+newCell->iYY)-radicalNew;
 		lMaxNew=0.5*(newCell->iXX+newCell->iYY)+radicalNew;
 		double ratio=lMinNew/lMaxNew;
-		if(ratio!=ratio || fabs(ratio)>1.0){
-			newCell->ecc=sqrt(1.0);
+		if(fabs(ratio)>1.0){
+			newCell->ecc=1.0;
 		}
 		else{
-			newCell->ecc=sqrt(1.0-ratio);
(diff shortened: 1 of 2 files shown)


JulianoGianlupi (approved):

CompuCell3D/core/CompuCell3D/plugins/MomentOfInertia/MomentOfInertiaPlugin.cpp
```diff
@@ -206,342 +186,97 @@ void CompuCell3D::MomentOfInertiaPlugin::field3DChange(const Point3D &pt, CellG
...
-		//	}else{
-
-		//		newCell->lX=0.0;
-		//		newCell->lY=newCell->iYY;
-		//	}
-
-		//}
-
+		auto c = cellOrientation_12(newCell->iXX, newCell->iYY, newCell->iXY);
+		newCell->ecc = (float)c[0];
```
CompuCell3D, line 191: I think we can introduce yet another function `assign_orientation_params(Cell Gcell, auto c)` where the 3 assignments below are done in this single function. The only complication would be dealing with lX, lY, lZ etc. So I will leave it as a optional item - if you feel like doing this extra step fine, if not, fine as well we can do it later

CompuCell3D/core/CompuCell3D/plugins/MomentOfInertia/MomentOfInertiaPlugin.cpp
```diff
@@ -206,342 +186,97 @@ void CompuCell3D::MomentOfInertiaPlugin::field3DChange(const Point3D &pt, CellG
...
-//
-//}
-
 vector<double> MomentOfInertiaPlugin::getSemiaxes(CellG *_cell)
 {
 
 	return (this->*getSemiaxesFctPtr)(_cell);
 }
 
 vector<double> MomentOfInertiaPlugin::getSemiaxesXY(CellG *_cell){
```
CompuCell3D, line 241: I think we are writing a bit complicated code by using iXY, iXZ etc instead of introducing inertia tensor where we could index axes by using numbers. We can think about refactoring it later. For now this is just an observation and perhaps a suggestion to add another item to our backlog

CompuCell3D (approved):
Many good chganges here . Looks good

PR #455, opened 2021-06-15
Feature/impl type

tjsego:
- CellType Plugin automatically assigns the lowest available integer to each type ID that is not specified via CC3DML. 
- Improvements throughout remove assumptions on the usage of cell type IDs, in particular that the set of type IDs does not have any particular structure (e.g., that the size of the set indicates the maximum ID).


CompuCell3D (approved):
Looks good - checked performance of ContactPlugin  and it is pretty much the same as before so I think we are ready to merge it

PR #457, opened 2021-06-15
Feature/source doc

tjsego:
- Automatically generates source code documentation from source code
- Adds documentation browsers to Twedit; documentation is asynchronously built by Twedit on demand
- Adds sphinx and pyqtwebkit to Twedit dependencies


cc3d/core/PySteppables.py
```diff
@@ -1776,11 +2155,15 @@ def invariantDistanceVectorBetweenCells(self, _cell_from, _cell_to):
 
     def invariant_distance_vector_between_cells(self, cell1, cell2):
         """
-        This function will calculate distance vector between  COM's of cells . Respects boundary conditions
-        :return: {ndarray} distance vector
+        This function will calculate distance vector between  COM's of cells. Respects boundary conditions
```
JulianoGianlupi, line 2158: This fcn says it respects BC, but `invariant_distance_vector` says it assumes periodic BC. I believe `invariant_distance_vector` does respect BC as it calls `CompuCell.BoundaryStrategy.getInstance()`

JulianoGianlupi (changes requested):
Clarify which distance methods respect BC and which don't

cc3d/core/PySteppables.py
tjsego: Addressed in https://github.com/CompuCell3D/CompuCell3D/pull/457/commits/238537aec7b36485a8268f7884ea9abdc0d80f0d

tjsego:
> Clarify which distance methods respect BC and which don't

This has been addressed in https://github.com/CompuCell3D/CompuCell3D/pull/457/commits/238537aec7b36485a8268f7884ea9abdc0d80f0d. Other clarifications in documentation should be made in separate PRs, as this PR is more concerned with generating documentation and making it available, rather than improving the content of the documentation itself.

JulianoGianlupi (approved):

cc3d/core/PySteppables.py
```diff
@@ -105,6 +116,29 @@ def __init__(self, plot_name, plot_type, attribute_name, function_obj=None):
...
+            <CellType TypeId="1" TypeName="Condensing"/>
+            <CellType TypeId="2" TypeName="NonCondensing"/>
+        </Plugin>
+
+    Usage in Python can be performed as follows,
+
+    .. code-block:: python
+
+        cell_type_fetcher: CellTypeFetcher
+        x: int = cell_type_fetcher.CONDENSING  # = 1
```
CompuCell3D, line 137: I forgot if we are capitalizing type names in cell_type_fetcher or not. Just something to double-check

CompuCell3D (approved):
I think it looks good. Nice that will have developer docs integrated with the actual code

PR #458, opened 2021-06-18
Feature/rem butils

tjsego:
This PR replaces the BasicUtils module with a new plugin manager, exception, random number generators and other minor features. 

Features
----------

- Added CC3DException for error handling. 
- Added ExtraMembers for dynamically adding members to class definitions. 
- Added RandomNumberGenerators with optional random number generators, including the current generator and Mersenne Twister. Mersenne Twister is the default, and the current generator is named "Legacy". 
- Promoted PluginManager to full implementation with new plugin approach.


CompuCell3D (approved):
Looks great

PR #481, opened 2021-08-13
Added boundary input validation against dimensions

tjsego:
Some plugins (e.g., NeighborTracker) don't behave as expected or even fail when specifying certain boundary conditions with a dimension of 1. This PR handles such cases by ignoring boundary condition inputs. Handling such cases can be converted to errors once #458 is approved and merged.


JulianoGianlupi (approved):

CompuCell3D (changes requested):
This is good but if we can throw exception and stop simulation this would be better. The reason is that this printout will be missed by most users and we will get tons of questions why CC3D does not work with periodic BS with dim=1 when they requested it. I think it is better to make people aware of this so simply Add the text you are printing in the exception and maybe say a bit more that when dimension == 1 then periodic BCs do not work and have users adjust their code. does it make sense?

tjsego:
> This is good but if we can throw exception and stop simulation this would be better. The reason is that this printout will be missed by most users and we will get tons of questions why CC3D does not work with periodic BS with dim=1 when they requested it. I think it is better to make people aware of this so simply Add the text you are printing in the exception and maybe say a bit more that when dimension == 1 then periodic BCs do not work and have users adjust their code. does it make sense?

Works for me. To avoid redundancy, I'll revise accordingly when #458 is merged.

CompuCell3D:
@tjsego  - rem_butils is merged so it is safe to update this PR by adding exception as we talked above

tjsego:
@CompuCell3D exceptions are implemented in 77edd32b43675cc629f739eeb70059b3fd6c458d.

CompuCell3D (approved):

Issue #476, opened 2021-08-07
SteadyStateDiffusionSolver can't edit the field in 3D excep at z=0


jpsluka:
SteadyStateDiffusionSolver documentation says you can poke values into a SteadyStateDiffusionSolver field, e.g., field[x,y,z]=value. This appears to only work if z=0 (as if the field is being treated as 2D and not 3D like SteadyStateDiffusionSolver is supposed to be)
[SteadyStateDiffusionSolver.zip](https://github.com/CompuCell3D/CompuCell3D/files/6949599/SteadyStateDiffusionSolver.zip)

CompuCell3D:
fixes merged

PR #484, opened 2021-08-15
Fixed a bug in SteadyStateDiffusionSolver field initialization - it w…

CompuCell3D:
…as setting field dimension z to be == 1

Addresses https://github.com/CompuCell3D/CompuCell3D/issues/476

Fixing hardcoded dim.z==1 for 3D SteadyStateDiffusion solver - even when user requested e.g. dim = (20, 20 20 3D) field it was internally set to be dim =20, 20, 1


CompuCell3D/CMakeLists.txt
```diff
@@ -29,10 +29,10 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL Windows)
     FIND_PACKAGE(PythonLibs)
     FIND_PACKAGE(PythonInterp)
     # INCLUDE(FindPythonLibs)
-    set(Python_EXECUTABLE ${PYTHON_EXECUTABLE} CACHE PATH "Python executable" FORCE)
```
tjsego, line 32: Edits to build scripts may cause conflicts with other current PRs. Are they necessary here?

CompuCell3D: This is what I am fighting all the time on my windows machine. If we keep FORCE argument I cannot configure Python in CMAKE GUI - it defaults to some weird paths. If I remove FORCE I can specify Python paths with no problem. Maybe you have a good solution for me @tjsego  so that I can keep FORCE arg in and get correct Python?

CompuCell3D: resolved the issue with addition new CMAKE option

CompuCell3D:
@tjsego @JulianoGianlupi : please take a looks at this PR - I resolved the issue with CMAKE FORCE modifier so I think we should be ready to merge.

JulianoGianlupi (approved):

tjsego (approved):

PR #489, opened 2021-08-17
Added simulation name validation in wizard

### LINKED ISSUE #465: Validate simulation name in Wizard
Certain characters in a simulation name can cause errors. These error-causing characters should be reported during simulation construction.

tjsego:
Wizard derives the name of a generated steppable from the simulation name, which puts constraints on valid simulation names. This PR addresses this issue by validating the simulation name and providing feedback to the user on error. 

Closes #465.

### DIFF
--- a/cc3d/twedit5/Plugins/CC3DProject/NewSimulationWizard.py
+++ b/cc3d/twedit5/Plugins/CC3DProject/NewSimulationWizard.py
@@ -894,6 +894,12 @@ class NewSimulationWizard(QWizard, ui_newsimulationwizard.Ui_NewSimulationWizard
                                     QMessageBox.Ok)
                 return False
 
+            elif not name.isidentifier():
+                QMessageBox.warning(self, "Invalid simulation name",
+                                    "Please specify a name with no spaces that begins with a letter or '_'",
+                                    QMessageBox.Ok)
+                return False
+
             else:
                 if directory != "":
                     self.plugin.configuration.setSetting("RecentNewProjectDir", directory)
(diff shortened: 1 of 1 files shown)


cc3d/twedit5/Plugins/CC3DProject/NewSimulationWizard.py
```diff
@@ -894,6 +894,12 @@ def validateCurrentPage(self):
                                     QMessageBox.Ok)
                 return False
 
+            elif not name.isidentifier():
+                QMessageBox.warning(self, "Invalid simulation name",
+                                    "Please specify a name with no spaces that begins with a letter or '_'",
```
JulianoGianlupi, line 899: with no spaces and no ```-```

JulianoGianlupi (changes requested):

cc3d/twedit5/Plugins/CC3DProject/NewSimulationWizard.py
CompuCell3D: you mean minus sign instead of underscore, right?

CompuCell3D (approved):
@tjsego if you can either change underscore to minus or confirm that current version is OK then we can merge it. I am approving it provisionally but please double-check this

cc3d/twedit5/Plugins/CC3DProject/NewSimulationWizard.py
JulianoGianlupi: yes

CompuCell3D (approved):
looks good

PR #494, opened 2021-08-27
Feature/rem player

tjsego:
Some staging here, to support migration of Player and Twedit. The build scripts currently have variables `INSTALL_PLAYER` and `INSTALL_TWEDIT` that collect the installation procedures to be migrated. Otherwise, this PR and #493 permit direct migration of Player and Twedit into separate packages that install directly into a core package as optional extensions.


CompuCell3D/core/post_install_tasks/windows/scriptSetup.py
```diff
@@ -1,171 +1,25 @@
...
-outFile.close()
-
-# Modifying celldraw run script
-
-import os.path
-
-if os.path.isfile("celldraw.bat.in.v2"):
-
-    # Modifying pifgenenrator file
+def localize_script(name: str):
```
CompuCell3D, line 9: great change. I dont know how many times I looked this previous script and wanted to refactor it but I always found something else to work on :)

cc3d/CompuCellSetup/SimulationThread.py
```diff
@@ -0,0 +1,127 @@
...
+            def f(CompuCell.Simulator, SimulationThread, CompuCell.SteppableRegistry) -> None
+
+        Note that persistent_globals.player_type is an optional global variable to use for conditionalizing which
+            function is returned here
+
+        See cc3d.CompuCellSetup.simulation_setup.main_loop for an example of how to set up an automated simulation run
+
+        :return: None
+        """
+        from cc3d.CompuCellSetup import main_loop
```
CompuCell3D, line 47: do we want this import here or at the top?

cc3d/core/Configuration/Configuration.py
```diff
@@ -0,0 +1,286 @@
+import os
```
CompuCell3D, line 1: general question: do we want to store settings in the cc3d/core or are settings part of the player. so far settings were part of the player and I think this is where they belong. Also Twedit has separate settings. My preference would be to keep settings close to the code that they belong to because if we separate Player and the core code then we will end up with player settings being in different repository that actual player. But maybe I am missing something here

cc3d/core/GraphicsOffScreen/GenericDrawer.py
```diff
@@ -34,7 +34,7 @@
 from .MVCDrawModel3D import MVCDrawModel3D
 from cc3d.core.GraphicsUtils.ScreenshotData import ScreenshotData
 from .Specs import ActorSpecs
-from cc3d.player5.Utilities.utils import extract_address_int_from_vtk_object
+from cc3d.core.GraphicsUtils.utils import extract_address_int_from_vtk_object
```
CompuCell3D, line 37: nice

cc3d/core/Configuration/Configuration.py
CompuCell3D: OK I see the reason. Off screen rendering uses configuration. ugh... I guess leaving configuration in cc3d/core is one form of solution. If that's the case than core code is a required dependency of the Player. I think it could work . we may want to think about expanding Configuration package with some form of introspection that would allow command line only users to view and modify settings without using PLayer. A simple , user friendly API for accessing configuration

cc3d/core/GraphicsOffScreen/primitives.py
```diff
@@ -0,0 +1,64 @@
+class Color:
```
CompuCell3D, line 1: nice

cc3d/player5/Plugins/ViewManagerPlugins/SimpleTabView.py
```diff
@@ -173,7 +168,7 @@ def __init__(self, parent):
 
         self.__outputDirectory = ""
 
-        self.__viewManagerType = "Regular"
+        self.__viewManagerType = ViewManagerType.REGULAR
```
CompuCell3D, line 171: i like enums here. Amazing how long we were tolerating plain strings :)

cc3d/player5/Simulation/main_loops.py
```diff
@@ -0,0 +1,139 @@
+import time
```
CompuCell3D, line 1: this is exactly where they belong

cc3d/player5/Utilities/__init__.py
```diff
@@ -3,4 +3,4 @@
 # __all__ = ['QVTKRenderWidget', 'QVTKRenderWindowInteractor']
 __all__ = ['QVTKRenderWidget']
 
-from . utils import *
\ No newline at end of file
+from cc3d.player5.Utilities.utils import *
```
CompuCell3D, line 6: minor comment but it would not be bad to eliminate `import  *` statements in favor of importing actual names used in a given module. this is more of a cleanup task so no need to rush now but something we can work as a side projects

cc3d/twedit5/Plugins/CC3DProject/unzipper.py
```diff
@@ -0,0 +1,91 @@
+import os.path
```
CompuCell3D, line 1: I know this file is duplicated and that's ok for now but I also wonder if perhaps this means that we need to created a fourth package (core_utils or something like that) that would be shared between core cc3d code, player, twedit++. core_utils would include functions like unzipper, configuration, and anything that is trully shared and does not depend on PyQt. This is just a thought

cc3d/twedit5/Plugins/PluginCCDProject.py
```diff
@@ -38,6 +38,8 @@
 from cc3d.twedit5.Plugins.CC3DProject.ParValDlg import ParValDlg
 from cc3d.twedit5.Plugins.CC3DProject.SerializerEdit import SerializerEdit
 from cc3d.twedit5.Plugins.CC3DProject.NewFileWizard import NewFileWizard
+from cc3d.twedit5.Plugins.CC3DProject.unzipper import Unzipper
+from cc3d.twedit5.Plugins import installed_player
```
CompuCell3D, line 42: like the change `installed_player`

CompuCell3D (changes requested):
It looks great. I like most of the changes . let me play with this branch and let's talk before we rubber stamp it and merge. But i like where this is going

CompuCell3D/CMakeLists.txt
```diff
@@ -187,6 +187,11 @@ INCLUDE(FindZLIB)
 OPTION(BUILD_PYINTERFACE "Build Python language wrappings" ON)
 
 OPTION(INSTALL_PLAYER "Move Player files to installation directory" ON)
+OPTION(INSTALL_TWEDIT "Move Twedit++ files to installation directory" ON)
+
+if(INSTALL_PLAYER OR INSTALL_TWEDIT)
+    set(BUILD_PYINTERFACE ON)
```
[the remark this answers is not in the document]
tjsego: `BUILD_PYINTERFACE` probably now is only relevant for controlling packaging specifics, and should always default to `TRUE`. For example, the conda package for GUIs will depend on a core conda package that was built with `BUILD_PYINTERFACE` set to `TRUE`. But I think in general it's a useful feature to be able to build locally with only the C++ libraries, for supporting pure C++ development. 

Also, what's built in this branch for `INSTALL_PLAYER` and `INSTALL_TWEDIT` will be migrated to separate repositories altogether. As such, the build scripts for the core package won't need defs for building GUIs _per se_. It's probably sufficient to just impose that whatever is built with `BUILD_PYINTERFACE` includes all supporting infrastructure for GUIs.

CompuCell3D/core/post_install_tasks/CMakeLists.txt
```diff
@@ -19,10 +19,8 @@ set(cc3d_py_source_dir ${CMAKE_SOURCE_DIR}/../cc3d)
 
 message("THIS IS cc3d_py_source_dir" ${cc3d_py_source_dir})
 
-if(INSTALL_PLAYER)
-#INSTALL(DIRECTORY  "@CMAKE_SOURCE_DIR@/${cc3d_py_dir}/CompuCellSetup" DESTINATION ${cc3d_py_dir_target}
+if(BUILD_PYINTERFACE)
```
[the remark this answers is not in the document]
tjsego: Not sure about this, since `BUILD_PYINTERFACE` also includes the Python API, which would probably be the intended user interface for the core package by itself.

cc3d/CompuCellSetup/SimulationThread.py
tjsego: I found placement here to be necessary to support implementations by external players, due to downstream declarations of global variables.

cc3d/player5/Utilities/__init__.py
tjsego: I always prefer to not use `*`. Makes my debugger too cluttered. I like the idea of doing what's suggested here throughout the codebase as a running cleanup task.

cc3d/twedit5/Plugins/CC3DProject/unzipper.py
tjsego: I'd had this same thought. Maybe the core package really should be just the computing, and a utilities package should contain all auxiliary functionality like configurations, zipping, etc.

CompuCell3D/CMakeLists.txt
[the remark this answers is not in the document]
CompuCell3D: ok I think this makes sense

cc3d/CompuCellSetup/SimulationThread.py
CompuCell3D: I thought it was something like this. Let's leave it for now and then we can work on writing "smart" import headers

JulianoGianlupi (approved):

cc3d/CompuCellSetup/SimulationThread.py
tjsego: Works for me.

CompuCell3D (approved):

PR #496, opened 2021-09-02
Added swig version spec to build scripts

tjsego:
Should help to resolve conflicts with multiple SWIG installs


CompuCell3D (approved):
Great. We needed this. On another note I wonder what it would take to move to swig 4.x and why swig broke backwards compatibility?

tjsego:
> Great. We needed this. On another note I wonder what it would take to move to swig 4.x and why swig broke backwards compatibility?

I think this could be fairly significant, but not too bad. SWIG 4.x eliminated `__swig_getmethods__` and `__swig_setmethods__`, which would require reworking the CellG python wrap and a few others. But I think it would be a nice upgrade.

PR #497, opened 2021-09-06
Feature/issue 459

CompuCell3D:
added more functionallity to cell color table - now users can change cell type colors and hide cell types in 3D directly from main window:

image


tjsego (approved):
Looks great.

PR #499, opened 2021-09-07
fix/run_script_cml_args

tjsego:
Adds more consistent handling of cml args through runScript.x compared to compucell3d.x.

### DIFF
--- a/cc3d/run_script.py
+++ b/cc3d/run_script.py
@@ -89,7 +89,10 @@ if __name__ == '__main__':
 
     rollbackImporter = RollbackImporter()
 
-    persistent_globals.simulation_file_name = cc3d_sim_fname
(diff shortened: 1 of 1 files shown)


JulianoGianlupi (approved):

cc3d/run_script.py
```diff
@@ -89,7 +89,10 @@ def handle_error():
 
     rollbackImporter = RollbackImporter()
 
-    persistent_globals.simulation_file_name = cc3d_sim_fname
+    current_dir = args.current_dir if args.current_dir else ''
+    cc3d_sim_fname_abs = join(current_dir, cc3d_sim_fname)
```
CompuCell3D, line 93: at some point we need to switch to pathlib.Path but this is a task for when we are bored :)

CompuCell3D (approved):

PR #504, opened 2021-10-14
Added support for python cell link inventory lists

tjsego:
Implements consistent iteration through link lists, whether retrieved from a link inventory or a cell link inventory.

### DIFF
--- a/cc3d/core/PySteppables.py
+++ b/cc3d/core/PySteppables.py
@@ -1540,7 +1540,7 @@ class SteppableBasePy(SteppablePy, SBMLSolverHelper, MaBoSSHelper):
         assert self.focal_point_plasticity_plugin is not None, 'Load focal point plasticity plugin'
         return self.focal_point_plasticity_plugin.getAnchorInventory()
 
-    def get_focal_point_plasticity_link_list(self):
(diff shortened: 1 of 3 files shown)


cc3d/core/PySteppables.py
```diff
@@ -1706,29 +1706,35 @@ def get_fpp_anchor_by_cell_and_id(self, cell: CompuCell.CellG, anchor_id: int) -
...
-    def get_fpp_internal_links_by_cell(self, _cell: CompuCell.CellG) -> CompuCell.FPPInternalLinkList:
+    def get_fpp_internal_links_by_cell(self, _cell: CompuCell.CellG) -> Optional[FocalPointPlasticityInternalLinkList]:
         """
         Get list of internal links by cell
         :param _cell: cell
         :return: internal links
         """
-        return self.fpp_internal_link_inventory.getCellLinkList(_cell)
+        if self.focal_point_plasticity_plugin is None:
+            return None
```
CompuCell3D, line 1726: Since many of those get_fpp_internal_links_xxx function are used in loop context we may consider raising StopIteration exception instead of returning None.

CompuCell3D (approved):
Looks good overall

PR #508, opened 2021-10-31
Feature/migrated pkgs fixes osx

CompuCell3D:
This PR fixes migrated_packages branch to compile properly on OSX and Windows (I assume it will carry over to linux too)


CompuCell3D/CMakeLists.txt
```diff
@@ -22,7 +22,16 @@ ENDIF (APPLE)
...
 
-FIND_PACKAGE (OpenMP)
+
+FIND_PACKAGE(OpenMP)
+IF(OPENMP_FOUND)
+message ("openmp c flags " "${OpenMP_C_FLAGS}")
+message ("openmp cxx flags " "${OpenMP_CXX_FLAGS}")
+SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS} ")
+#SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS} -std=c++11 -stdlib=libc++")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS} -std=c++11 -stdlib=libc++")
```
tjsego, line 32: Fails with gcc. See fix in https://github.com/CompuCell3D/CompuCell3D/pull/509.

CompuCell3D/core/pyinterface/PlayerPythonNew/CMakeLists.txt
```diff
@@ -59,9 +59,13 @@ target_link_libraries(FieldExtractor ${LIBS})
 
 install(TARGETS FieldExtractor
         EXPORT FieldExtractor
-        RUNTIME DESTINATION ${COMPUCELL3D_INSTALL_BIN_DIR}
-        LIBRARY DESTINATION ${COMPUCELL3D_INSTALL_LIB_DIR}
-        ARCHIVE DESTINATION ${COMPUCELL3D_INSTALL_LIB_DIR}
+        RUNTIME DESTINATION bin
```
tjsego, line 62: These changes are fine, though I suspect that they create some redundancy in the main module __init__. These nested /bin and /lib paths are currently append on import of cc3d. Should probably verify whether this procedure can be removed due to this proposed change and remove it if so.

tjsego: Additionally, if using the build variables ``COMPUCELL3D_INSTALL_BIN_DIR`` and ``COMPUCELL3D_INSTALL_LIB_DIR`` is eliminated by this PR, then can we remove them altogether in this PR?

conda-recipes-compucell3d/meta.yaml
```diff
@@ -23,8 +23,8 @@ build:
 requirements:
```
tjsego, line 23: Not sure about including this second recipe here. On Linux, conda build tries to chain together building both recipes, and then fails due to missing dependencies for this metapackage. Obviously, this then makes it impossible to build all packages in one automated workflow, since building one of these recipes automatically triggers building the other. I don't see this behavior on Windows, but also haven't found any way to disable it on Linux. 

Does it make more sense to have a separate repo just for storing this recipe? Something like a repo for a simple conda-forge recipe, maybe. Or maybe we could just post the metapackage on conda-forge? Let's discuss.

conda-recipes/meta.yaml
```diff
@@ -27,22 +27,33 @@ requirements:
     - vtk=8.2
     - eigen
     - tbb-devel=2020
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
```
tjsego, line 31: These are redundant. Proposing to remove them in https://github.com/CompuCell3D/CompuCell3D/pull/509.

conda-recipes/meta.yaml
```diff
@@ -27,22 +27,33 @@ requirements:
...
+    - numpy=1.18
+    - vtk=8.2
     - {{ pin_compatible('numpy') }}
     - {{ pin_compatible('vtk') }}
     - scipy
     - pandas
     - jinja2
     - deprecated
     - psutil
+    - cc3dext_maboss
```
tjsego, line 52: This can't be here, since it will create a dependency loop. Furthermore, if I understand correctly, then the extensions are optional features. 

Technically, it's ok for now. But once the ``cc3d`` conda package is published, it will become a dependency of ``cc3dext_maboss``. Proposing to remove it in https://github.com/CompuCell3D/CompuCell3D/pull/509.

tjsego (changes requested):

conda-recipes-compucell3d/meta.yaml
CompuCell3D: We can definetely introduce separate repository with compucell3d meta recipe. I added it here as a placeholder . BTW they way I am building package I do:

`cd conda-recipes`
`conda build .`

this way you are forcing conda to use recipe from current dir. But I have not tested it on Linux. Does conda beahve differently on Linux?

CompuCell3D/CMakeLists.txt
CompuCell3D: I am sure it does. I remember breaking things to get it working on OSX. Forgot to clean up :)

CompuCell3D/core/pyinterface/PlayerPythonNew/CMakeLists.txt
CompuCell3D: I would be in favor of eliminating variables we dont use or we can replace by simpler solutions

conda-recipes-compucell3d/meta.yaml
tjsego: This is the same procedure that I use both on Windows and Linux, so maybe it's a conda issue.

tjsego (approved):
Looks good. We can clean up there rest in future PRs.

PR #513, opened 2021-12-21
Refactored code to remove false positive AVG report coming from lengt…

CompuCell3D:
…h constraint dll. What was required was to declare most variables at the top of the function and not in a JIT fashion during definition of the variable

This is  very minor refactoring to stop AVG to reporting malware (false positive) from length constrain plugin that.


tjsego (approved):
Interesting that this fixes this issue. Thanks for hunting it down. I'd like to understand this well enough to prevent this issue in future dev.

CompuCell3D:
@tjsego the problem is that nobody (except AVG) people understands the issue. Their algorithms have a lot of proprietary heuristics that sometimes act weird.

PR #540, opened 2022-05-24
Fix/issue 536

### LINKED ISSUE #536: Running cc3d 4.3.0 headless through external python scripts crashes with fpp links
Simulation runs fine through player however throws an error shortly after step 0 while running a simulation with FPP links.

Here is the error:
image

I am running cc3d 4.3.0 on Windows 11 installed through conda forge into a conda environment.

tjsego:
Fixes #536. Probably, the pattern in this fix should be implemented throughout.

### DIFF
--- a/cc3d/core/BasicSimulationData.py
+++ b/cc3d/core/BasicSimulationData.py
@@ -1,10 +1,36 @@
-import sys, os
+from weakref import ref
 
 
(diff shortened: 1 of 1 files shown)


CompuCell3D (approved):
Looks good to me

PR #547, opened 2022-06-01
Added Python API pipeline to steppable interface

tjsego:
* Added minor convenience properties to steppable interface for easier access to data pipeline via the Python API
* Updated CallableCC3D demos
* Removed environment setup scripts, since they're obsolete as of v4.3.0.


CompuCell3D (approved):
Looks good!

PR #548, opened 2022-06-07
Cell inventory watcher

tjsego:
Added an interface for clients to receive notifications of change to the cell inventory, with support in Python.


CompuCell3D (approved):
Loogs good

PR #551, opened 2022-06-08
Code documentation, OpenMP implementation added for Reaction-Diffusion Finite Volume Solvers

DeconvFFT:
- This version adds documentation for Reaction-Diffusion Finite Volume Solvers inside PDE solvers and adds OpenMP support for parallel implementation in Reaction-Diffusion Finite Volume Solvers. Using OpenMP enables cross-platform support for parallel implementation which was lacking for the Parallel Patterns Library.


tjsego (changes requested):
Can you remove all files associated with the build and VS code?

DeconvFFT:
@tjsego, removed build and .vscode directories

tjsego:
@DeconvFFT there are still build-related files. Notice how the PR is ~224k lines of additional code. This needs to be reduced to only modifications to the actual source code.

CompuCell3D:
My suggestion is to keep source code directory  separate from cmake-generated binary dir. I.e. if you clone cc3d repo to e,g, c:\cc3d_git then during cmake configuration pick binary directory to be e.g.  c::\cc3d_git_build. This way you will never have to deal with situation where you end up with additional files in the git repo

PR #552, opened 2022-06-13
Code documentation, OpenMP implementation added for Reaction-Diffusion Finite Volume Solvers

DeconvFFT:
This version adds documentation for Reaction-Diffusion Finite Volume Solvers inside PDE solvers and adds OpenMP support for parallel implementation in Reaction-Diffusion Finite Volume Solvers. Using OpenMP enables cross-platform support for parallel implementation which was lacking for the Parallel Patterns Library.


DeconvFFT:
@tjsego, I was able to remove build-related files.

tjsego (changes requested):
Much better. Looks like there are some lingering files from installing into the source directory. Please remove the following from the changes: 

- GPUEnabled.h
- include/*
- post_install_tasks/*

Also, please restore core/CompuCell3D/plugins/MomentOfInertia/Makefile to an unchanged state.

DeconvFFT:
@tjsego, wrapped print statements in debug macro

tjsego (approved):

PR #554, opened 2022-06-21
Added preliminary support for Python>3.7 on Windows

tjsego:
Not tested against newer versions of Python, but this PR includes upgrades that are known to be necessary for loading dlls at runtime on Windows according to changes in the Python API in version 3.8.


CompuCell3D (approved):
I think this looks OK. We can fine-tune the code later but as it looks now it should work with 3.7 adn possibly with 3.8+

PR #559, opened 2022-07-08
Minor fix in FieldExtractor

tjsego:


CompuCell3D (approved):
Wow , you have great eye. I think this really might be the issue

PR #563, opened 2022-07-14
Fixed python api autodoc dependency search

tjsego:
Wouldn't hurt to add dependencies here to the conda package meta, too, but I think it's fine for now if not


CompuCell3D (approved):
Looks good

PR #580, opened 2022-09-02
Feature/cpython

tjsego:
* Adds base class ``CC3DPyObject`` to generalize python class features in C++ (*e.g.*, ``CellG.getdict``) that are currently implemented in swig layer
* Implements Python dictionary on ``CellG`` and link definitions
* Adds Python support for debug builds on windows (still missing at least *tbb_debug.lib* on builds using conda environments)

Note that minor edits in ``RollbackImporter`` were not expected, and why changes otherwise required them is unclear to me. 

Note that this PR deprecates minor features that are currently used to support attachment of Python dictionaries to ``CellG`` instances. A deprecation notice should be added or they should be outright removed, pending approval of this PR.


JulianoGianlupi:
Looks good to me. One question, why not put the deprecation warning in this PR?

tjsego:
> Looks good to me. One question, why not put the deprecation warning in this PR?

Good question. I'm looking to see how everyone would like to proceed. I'm happy to implement notices or remove them in this PR if that's the consensus.

CompuCell3D/core/pyinterface/FiPyInterface/FiPyInterface.cpp
```diff
@@ -1,7 +1,7 @@
 #include <iostream>
```
CompuCell3D, line 1: I would not mind getting rid of this module (FiPy) . It was experimental a while ago and I don't think anybody uses it

cc3d/core/RollbackImporter.py
```diff
@@ -70,7 +70,10 @@ def _import(self, name, globals=None, locals=None, fromlist=[], level=0):
```
CompuCell3D, line 70: We need to figure out how to structure the code so that we don't have to use RollbackImporter. currently it unloaded all swig modules and allows us to run multiple simulations without closing the player but maybe there is a better way. RollbackImporter is a bit complicated

CompuCell3D (changes requested):
I think it looks OK. Since this is quasi-major change I would like to do a test-drive before approving it. I would add deprecation warnings in the right places so that we can see what is affected and if there are any breaking changes

For now I am marking it as "Request changes" but this is really my way of saying let me test it first. NO major changes - except for deprecation warnings are requested

cc3d/core/RollbackImporter.py
tjsego: Agreed. A service architecture like what we're doing in [feature/jupyter](https://github.com/CompuCell3D/CompuCell3D/tree/feature/jupyter) might do the trick.

CompuCell3D:
ok so it looks like CC3D crashes with this PR on OSX.  Running ` python -m cc3d.run_script -i /Users/m/src/Demos-local/Models/cellsort/cellsort_2D/cellsort_2D.cc3d -f 1000`
results in this error:


`itDim=(20,20,1)
Fatal Python error: PyMUTEX_LOCK(_PyRuntime.ceval.gil.mutex) failed

zsh: abort      python -m cc3d.run_script -i  -f 1000`

We can either fix it or put this PR on hold and focus on more pressing issues. I will leave it up to you to decide 

BTW I was compiling this on M1 mac in x86 mode

tjsego:
Holding off on this for more pressing issue. Let's revisit when more opportune.

PR #592, opened 2022-10-27
Added support for SecretionData in RDFVM solver

tjsego:


JulianoGianlupi (approved):

CompuCell3D/core/CompuCell3D/steppables/PDESolvers/ReactionDiffusionSolverFVM.cpp
```diff
@@ -54,8 +55,10 @@ ReactionDiffusionSolverFVM::~ReactionDiffusionSolverFVM()
 	delete lockPtr;
 	lockPtr = 0;
 
-	delete fvMaxStableTimeSteps;
-	fvMaxStableTimeSteps = 0;
+	if(fvMaxStableTimeSteps) {
```
CompuCell3D, line 58: I think we should be able to write a code without explicit memory allocation or use shared_ptr https://en.cppreference.com/w/cpp/memory/shared_ptr

the only downside of shared pointers is run time performance but I doubt this will be an issue here. we can fix it in the next release I am just making a comment here.

CompuCell3D/core/CompuCell3D/steppables/PDESolvers/ReactionDiffusionSolverFVM.cpp
```diff
@@ -54,8 +55,10 @@ ReactionDiffusionSolverFVM::~ReactionDiffusionSolverFVM()
 	delete lockPtr;
```
CompuCell3D, line 55: I would also try using shared pointer instead of explicitly managing memory

CompuCell3D/core/CompuCell3D/steppables/PDESolvers/ReactionDiffusionSolverFVM.h
```diff
@@ -8,6 +8,7 @@
 
 #include <CompuCell3D/plugins/PixelTracker/PixelTracker.h>
 #include <CompuCell3D/plugins/PixelTracker/PixelTrackerPlugin.h>
+#include <CompuCell3D/plugins/NeighborTracker/NeighborTrackerPlugin.h>
```
CompuCell3D, line 11: I think we may declare NeighborTrackingPlugin here:
namespace CompuCell3D{
   class NeighborTrackingPlugin
   ...

and we can move include to .cpp file  - minor issue though

CompuCell3D/core/CompuCell3D/steppables/PDESolvers/ReactionDiffusionSolverFVM.h
```diff
@@ -1159,6 +1162,16 @@ namespace CompuCell3D {
 		 * @return false 
 		 */
 		bool isValid(Point3D pt) { try { pt2ind(pt); return true; } catch (CC3DException) { return false; } }
+
```
CompuCell3D, line 1165: maybe we should try adding c++ docstrings to other places in the code? I like it

CompuCell3D/core/CompuCell3D/steppables/PDESolvers/ReactionDiffusionSolverFVM.cpp
```diff
@@ -1554,6 +1680,28 @@ double ReactionDiffusionSolverFV::getFieldDiffusivityInMedium(unsigned int _fiel
 	else { return solver->getCellDiffusivityCoefficient(cell, _fieldIndex); }
 }
 
+double ReactionDiffusionSolverFV::cellInterfaceFlux(unsigned int _fieldIndex, unsigned int _surfaceIndex, CellG *cell, CellG *nCell) {
```
CompuCell3D, line 1683: Does it mean that we will be able to specify what is going on at the cell interface? If so many people will be happy

CompuCell3D (approved):
Looks good - some minor comments that we can address in the next iterations

CompuCell3D/core/CompuCell3D/steppables/PDESolvers/ReactionDiffusionSolverFVM.cpp
tjsego: Correct. It's currently available on the API, and support needs added in the CC3DML.

Issue #575, opened 2022-08-10
[Major] When changing cell types with different diffusion constant the diffusion constant is not updated


JulianoGianlupi:
If the cell is frozen, when we change from a cell type with diffusion constant=0 to a cell type with D>0 the chemical does not diffuse through the D>0 cell. I've attached a simulation that exemplifies the problem

[debug_diffusion.zip](https://github.com/CompuCell3D/CompuCell3D/files/9304260/debug_diffusion.zip)

This bug happens because the D update happens only on successful pixel flips. This sim exemplifies that the diffusion only starts with pixels that have changed cell

[debug_diffusion_unfrozen.zip](https://github.com/CompuCell3D/CompuCell3D/files/9304277/debug_diffusion_unfrozen.zip)

pdalcastel:
I will try:

1) Delete the cell
2) Build a new cell of a different type over the same pixels

And let you know

maciekswat:
fixed. thanks @JulianoGianlupi  for reporting it

PR #607, opened 2022-12-03
Implemented generation of cell type and cell id indicator arrays righ…

maciekswat:
…t before calling the step function of the steppable. It fixes the staleness of those arrays reported in https://github.com/CompuCell3D/CompuCell3D/issues/575. As of now cell type and cell_id arrays reflect the most recent configuration of the lattice regardless of whether we are using frozen cells or not. The overhead associated with update is small


tjsego (approved):
Looks good

PR #631, opened 2023-03-10
Minor fix for vtk 9

tjsego:


maciekswat (approved):
I was about to start looking at that. Glad you found the fix!

PR #634, opened 2023-03-24
adding two aliases for `SteppableBasePy.shared_steppable_vars` 

JulianoGianlupi:
based on James' request

### DIFF
--- a/cc3d/core/PySteppables.py
+++ b/cc3d/core/PySteppables.py
@@ -403,6 +403,16 @@ class SteppableBasePy(SteppablePy, SBMLSolverHelper, MaBoSSHelper):
                                          'lambdaVecY', 'lambdaVecZ', 'fluctAmpl']
 
     @property
+    def shared_steppable_dict(self):
+        """Alias for self.shared_steppable_vars"""
+        return self.shared_steppable_vars
+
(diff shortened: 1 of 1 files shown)


JulianoGianlupi:
I tested adding keys to the dictionary and aliased dictionary and it worked

cc3d/core/PySteppables.py
```diff
@@ -402,6 +402,16 @@ def __init__(self, *args, **kwds):
                                          'targetClusterSurface', 'lambdaClusterSurface', 'type', 'lambdaVecX',
                                          'lambdaVecY', 'lambdaVecZ', 'fluctAmpl']
 
+    @property
+    def shared_steppable_dict(self):
+        """Alias for self.shared_steppable_vars"""
+        return self.shared_steppable_vars
+
+    @property
+    def dict(self):
```
maciekswat, line 411: I would remove dict function because it might be confusing (too generic). shared_steppable_dict is a better name for the new property. @JulianoGianlupi , ake this change and I will merge the PR

maciekswat (changes requested):
minor change requested

maciekswat (approved):
Perfect!

PR #644, opened 2023-04-27
Fixed missing data in screenshot output

tjsego:
Screenshot specs aren't respected through Python API without this fix


maciekswat (approved):
Good catch!

PR #650, opened 2023-06-09
Minor fixes for MSVC 2019 and python init

tjsego:
I can build with MSVC 2019 using these fixes. Metadata for conda-build would need updated if we upgrade the build pipeline for windows.


maciekswat (approved):
I think this looks good.

PR #664, opened 2023-10-01
Fixed initial vizualization in pure Python

tjsego:


cc3d/core/GraphicsUtils/CC3DPyGraphicsFrame.py
```diff
@@ -348,6 +348,9 @@ def __init__(self,
     def start(self, auto_shutdown: bool = True):
         """Start the event loop"""
 
+        self.draw()
```
maciekswat, line 351: does this work fine in the PLayer as well?

tjsego: No effect in player. The player implementation doesn't derive from this class.

maciekswat: perfect!

maciekswat (approved):
approved

PR #665, opened 2023-10-01
Added rtd build support for Python API docs

tjsego:
Supports new readthedocs-hosted Python API docs. We need these docs to support both development and new all-Python specification.


maciekswat (approved):
nice!

PR #675, opened 2023-10-23
Fixed library loading on windows

tjsego:


tjsego:
This is probably only necessarily for builds from source.

CompuCell3D (commented):
This is interesting because I removed those lines precisely because local builds from source were not working properly. Let's meet and talk this one through - or maybe let's add a variable , say `LOCAL_BUILD` that enables this part conditionally. What do you think?

PR #681, opened 2023-11-14
Added Movie Feature 

### LINKED ISSUE #669: Code to generate movies

fyffep:
See also: https://github.com/CompuCell3D/cc3d-player5/pull/40
Closes #669.


maciekswat:
@fyffep let's make this PR go to develop branch, not master. You should be able to use Edit button - top, right corner to make this change

maciekswat (approved):
looks very good. The only thing I would do is to move function `makeMoviesWithSettings` to player code (because it imports CC3D Player - hence it does not belong to this repository) and replace it with another function that does not rely on player settings but instead takes those settings as arguments - it looks like `makeMovie` is such function

fyffep:
> looks very good. The only thing I would do is to move function `makeMoviesWithSettings` to player code (because it imports CC3D Player - hence it does not belong to this repository) and replace it with another function that does not rely on player settings but instead takes those settings as arguments - it looks like `makeMovie` is such function

I think it keeps the code simple to have everything in one file since it's currently a small module. I understand your point, but aren't Player and the core code already tightly coupled?

maciekswat:
> > looks very good. The only thing I would do is to move function `makeMoviesWithSettings` to player code (because it imports CC3D Player - hence it does not belong to this repository) and replace it with another function that does not rely on player settings but instead takes those settings as arguments - it looks like `makeMovie` is such function
> 
> I think it keeps the code simple to have everything in one file since it's currently a small module. I understand your point, but aren't Player and the core code already tightly coupled?

They are coupled to some extent. but if you run CC3D without player (as some teams do) you do not have access to Settings or Configuration. And we did some coding to make sure that you can run CC3D without requiring PyQt. So from this perspective, it would be nice to have a function that makes movies and operates in the complete absence of Configuration or any Player-related code or data structures (this function would live in core compucell3d code) and then have a thin wrapper around it that lives in the PLayer code but uses Configuration. Does this make sense?

PR #686, opened 2023-11-22
Added Text Option to Movie Maker; Fixed Delay in Movies 

### LINKED ISSUE #684: move function `makeMoviesWithSettings` to player
Let's move function `makeMoviesWithSettings` to player code (because it imports CC3D Player - hence it does not belong to this repository) and replace it with another function that does not rely on player settings but instead takes those settings as arguments - it looks like `makeMovie` is such function

fyffep:
This goes with the PR https://github.com/CompuCell3D/cc3d-player5/pull/42

Resolves #684 and #685 and #669


cc3d/core/GraphicsUtils/MovieCreator.py
```diff
@@ -2,16 +2,20 @@
 import os
 import subprocess
 import tempfile
-import shutil
 
-def makeMovie(simulationPath, frameRate, quality):
+
+def makeMovie(simulationPath, frameRate, quality, enableDrawingMCS=True):
```
maciekswat, line 7: The frames are out of order and to fix it you need to change line 42 from 
```python
 for fileNameExt in os.listdir(inputPath):
```

```python
 for fileNameExt in sorted(os.listdir(inputPath)):
```

maciekswat (changes requested):
Looks mostly good except one change that needs to restore proper order of the frames. 



I would also replace all references to `os.path` with pathlib equivalents. We stopped using os.path a while ago and it would be good to use newer pathlib  - I will add an issue to resolve this later. For now let's merge the code as is (with the change I outlined)

PR #687, opened 2023-11-23
Feature/developer zone fix

maciekswat:
Fixes to Cmake files to get DeveloperZone to compile on Windows


fyffep:
I wanted to test this branch. How can I load in the files that were changed? They are not present in my Miniconda environment. Similarly, Visual Studio gives an irrelevant compile error when I try to run INSTALL_PROJECTS.

maciekswat:
To test it you use your standard binary install of CC3D the one that is typically in c:\compucell3d. Then you follow the guidelines displayed in the developer zone configuration dialog: https://compucell3ddevelopersmanual.readthedocs.io/en/master/configuring_developer_zone.html

It would be actually good to see if you can reproduce those steps on your machine and if something is outdated we could update the documentation. But normally you clone this branch to a separate location and then open Twedit from c:\Compucell3d and if you follow the guidelines it should work. LEt me know if it works for you

PR #691, opened 2023-12-01
Fixed sim_input name class in Python API

tjsego:
Name clash is crashing some features of `CC3DSimService`.


maciekswat (approved):
Great!

PR #693, opened 2023-12-01
Add on-demand lattice snapshots to Python API

tjsego:
Doesn't seem to affect Player, but additional testing would be appreciated.


maciekswat (approved):
Looks good. I checked out the branch and the code runs fine. The only thing I would to is to add a demo showing how to use store_lattice_snapshot

Issue #652, opened 2023-06-30
Gillespie replicates are synchronized, no abitliy to set a seed


jpsluka:
When using SBML and the Gillespie solver all replicates are synchronized. here is no method to set a seed for each model instance or to otherwise knock the simulator out of synch.

As a hack, the models can be stepped a large number of times after each model is loaded into a cell.

maciekswat:
Jim, can you post minimal example of simulation using Gillespie solvers? It would help me or TJ to quickly jump ion and fix the issue

jpsluka:
[Gillespie_random_number.zip](https://github.com/CompuCell3D/CompuCell3D/files/12399020/Gillespie_random_number.zip)

The model has two cells each with a simple A->B ODE and runs using Gillespie. The two cell's [A] are synchronized, which they should not be. I don't see any way to assign random number seeds in the model spec, though that should not be needed, different instances of the same model should already be using different random numbers.

jpsluka:
CC3D release candidate 4.5.0 appears to still have this bug.

tjsego:
The seed on a RoadRunner instance can be accessed with `rr.integrator.seed`. CC3D doesn't set or randomize the seed of newly created RoadRunner instances by default, though adding an option to do so in the various constructors isn't a bad idea.

jpsluka:
So based on TJ's comment, to load an antimony string, then set to Gillespie and change the random number seed:

```
        ### Antimony model string
        model_string = '''        
        //Equations
        A -> B;  k1*A
        //Parameters        
        k1 = 0.02;
        //Initial Conditions
        A = 100
        B = 0
        '''  
        
        ### create cell 
        newCell1 = self.new_cell(self.ACELL)
        self.cell_field[45:50, 45:50, 0] = newCell1
        newCell1.targetVolume = newCell1.volume
        newCell1.lambdaVolume = 10.

        ### add Antimony to cell
        RRinst1 = self.add_antimony_to_cell(model_string=model_string, model_name='firstOrder', \
                                            cell=newCell1, step_size=1, integrator='gillespie' )
        ### Change random seed used in Gillespie
        newCell1.sbml._rr_firstOrder.integrator.seed = 123
       
```
Note how the model_name (**firstOrder**) propagates.

jpsluka:
[...]

PR #698, opened 2024-01-06
Added a fix that by default randomizes seed for Gillespie solvers but…

maciekswat:
… also allows runs with fixed seed via

e.g. self.set_gillespie_integrator_seed(20).

Added example of using Gillespie solver with fixed random seed and, the default setting, fully randomized seed

Addresses https://github.com/CompuCell3D/CompuCell3D/issues/652

### DIFF
--- a/cc3d/core/SBMLSolverHelper.py
+++ b/cc3d/core/SBMLSolverHelper.py
@@ -5,6 +5,9 @@ import types
 from cc3d.cpp import CompuCell
 from cc3d import CompuCellSetup
 from deprecated import deprecated
+from random import randint
+
+
 
 # Test for Antimony installation
 try:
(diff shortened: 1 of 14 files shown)


cc3d/core/SBMLSolverHelper.py
```diff
@@ -111,6 +115,36 @@ def __default_mutable_type(self, obj: Union[object, list, dict], obj_default: Un
...
+            return None
+
+    def get_gillespie_integrator_max_seed(self) -> int:
+
+        try:
+            return CompuCellSetup.persistent_globals.gillespie_integrator_max_seed
+        except AttributeError:
+            return int(2e9)
+
+    def set_integrator_seed(self, integrator_name: str, rr: RoadRunnerPy):
```
tjsego, line 138: Gillespie is the only valid integrator for setting a seed, so is there any reason to require an integrator name as argument?

maciekswat: I think so. In case roadrunner team comes up with another algorithm e.g. gillespie_2. it is more of future-proofing - if such thing even exists ...

tjsego (approved):

PR #702, opened 2024-01-21
Feature/surface energy neighbor order

maciekswat:
This is the fix to the surface plugin that allows us to easily specify how we want to compute surface (meaning if we want to go beyond first nearest neighbor). Previously we had to use the following (in this order of plugins):

```xml
  <Plugin Name="SurfaceTracker">
        <MaxNeighborOrder>4</MaxNeighborOrder>
    </Plugin>

    <Plugin Name="Surface">
        <TargetSurface>120</TargetSurface>
        <LambdaSurface>0.5</LambdaSurface>
    </Plugin>

```

and the new way is much simpler:
```xml
    <Plugin Name="Surface">
        <TargetSurface>120</TargetSurface>
        <LambdaSurface>0.5</LambdaSurface>
        <NeighborOrder>4</NeighborOrder>
    </Plugin>
```


fyffep:
It's a sensible change, and it works on my machine. I have to remind myself to update the docs for this.

fyffep (approved):

maciekswat:
@fyffep  If you can update the docs that would be awsome

PR #710, opened 2024-03-11
Fixes/improvements to Python API visualization

tjsego:
* Moves default settings to XML-based with single file load per visualization process (sqlite was too problematic for multiple processes)
* Fixes extra field visualization

### DIFF
--- a/cc3d/core/Configuration/SettingUtils.py
+++ b/cc3d/core/Configuration/SettingUtils.py
@@ -108,9 +108,9 @@ def _default_setting_path_xml() -> str:
     Returns the path to the default settings xml data
     """
     if sys.platform.startswith('darwin'):
-        return os.path.join(_default_setting_dir(), '_settings_osx.xml')
+        return os.path.join(_default_setting_dir(), SETTINGS_XML_OSX_FILE_NAME)
     else:
-        return os.path.join(_default_setting_dir(), '_settings.xml')
+        return os.path.join(_default_setting_dir(), SETTINGS_XML_FILE_NAME)
 
 
 def _global_settings_dir():
--- a/cc3d/core/Configuration/SettingUtils.py
+++ b/cc3d/core/Configuration/SettingUtils.py
@@ -1,8 +1,12 @@
 import os
 import shutil
 import sys
+from typing import Any, Dict
+import warnings
+from xml.etree import ElementTree
 
 from cc3d.core.DefaultSettingsData import *
+from .settingdict import SerializerUtil
 
 
 def load_settings(setting_path):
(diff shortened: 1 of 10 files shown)


cc3d/core/Configuration/SettingUtils.py
```diff
@@ -212,3 +231,51 @@ def synchronize_global_and_default_settings(default_settings, global_settings, g
...
+        return '0'
+    return _val
+
+
+def default_settings_dict_xml() -> Dict[str, Any]:
+    """
+    Returns default data dictionary
+    """
+    fp = _default_setting_path_xml()
+    if not os.path.isfile(fp):
```
fyffep, line 252: @maciekswat said something about wanting to use pathlib instead of the OS module, but I forgot the reason.

cc3d/core/Configuration/SettingUtils.py
```diff
@@ -212,3 +231,51 @@ def synchronize_global_and_default_settings(default_settings, global_settings, g
...
+    sz_util = SerializerUtil()
+
+    def _import_data(_el):
+        _val_text = _handle_settings_value_xml(_el.text)
+        _type_fcn = sz_util.type_2_deserializer_dict[_el.attrib['Type']]
+        return _el.attrib['Name'], _type_fcn(_val_text)
+
+    result = {}
+    for el in data_root.findall('Settings'):
+        for el_e in el.findall('e'):
```
fyffep, line 266: nit: I might add a comment to reveal that `e` is the name of the XML tag

cc3d/core/Configuration/SettingUtils.py
```diff
@@ -212,3 +231,51 @@ def synchronize_global_and_default_settings(default_settings, global_settings, g
...
+                    try:
+                        result_e.__setitem__(*_import_data(el_e_e))
+                    except (KeyError, TypeError):
+                        pass
+                result[el_e.attrib['Name']] = result_e
+            else:
+                try:
+                    result.__setitem__(*_import_data(el_e))
+                except (KeyError, TypeError):
+                    pass
```
fyffep, line 279: Should these `except` blocks print something like, "The setting for ___ could not be read"?

fyffep (approved):

cc3d/core/Configuration/SettingUtils.py
maciekswat: unless you "normalize" paths - meaning use forward or reverse slashes depending on the platform the os may mix forward and reverse slashes. pathlib handles it much better.  In general, pathlib gives you more path-centric  access to all file operations. with os modules, you need a lot more modules to handle certain operations e.g. making directly

maciekswat (commented):
Few questions: 
1. does it replace sqlite settings? 
2. If so how do you handle synchronization? sqllite is not perfect but we ppicked it because with xml settings we had all sorts of problems mainly due to synchronization. I am wondering if there are standard solutions for storing setting (except Windows registry which is not what we want here )

cc3d/core/Configuration/SettingUtils.py
tjsego: I don't see pathlib being used in this neighborhood of the code, so can we reserve pathlib utlilization for a separate PR?

tjsego:
> Few questions:
> 
> 1. does it replace sqlite settings?
> 2. If so how do you handle synchronization? sqllite is not perfect but we ppicked it because with xml settings we had all sorts of problems mainly due to synchronization. I am wondering if there are standard solutions for storing setting (except Windows registry which is not what we want here )

1. No. It avoids sqlite altogether. 
2. No synchronization is performed across runs or even windows. I opted to initialize every new graphics frame in the Python API from defaults to ensure reproducible visualization.

cc3d/core/Configuration/SettingUtils.py
```diff
@@ -78,20 +82,35 @@ def _load_sql_settings(setting_path):
...
+
+
+def _default_setting_path_xml() -> str:
+    """
+    Returns the path to the default settings xml data
+    """
+    if sys.platform.startswith('darwin'):
+        return os.path.join(_default_setting_dir(), '_settings_osx.xml')
+    else:
+        return os.path.join(_default_setting_dir(), '_settings.xml')
```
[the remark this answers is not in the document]
tjsego: Addressed in https://github.com/CompuCell3D/CompuCell3D/pull/710/commits/c18f074886a384f8413103e460840cbbe14bcda5.

cc3d/core/Configuration/SettingUtils.py
tjsego: Addressed in https://github.com/CompuCell3D/CompuCell3D/pull/710/commits/c592f764ecd9cdd9bf186665fe93830480a1090c.

cc3d/core/Configuration/SettingUtils.py
tjsego: Agreed. Addressed in https://github.com/CompuCell3D/CompuCell3D/pull/710/commits/c592f764ecd9cdd9bf186665fe93830480a1090c.

cc3d/core/Configuration/SettingUtils.py
fyffep: Definitely. I have done this previously.

maciekswat (approved):
Looks good. Tested on  OSX and it was just fine!

Issue #712, opened 2024-04-02
Player Crashes when opening DML file


maciekswat:
https://www.reddit.com/r/CompuCell3D/comments/1bu3y3w/player_crashes_while_opening_dml_files/

maciekswat:
fixed

PR #713, opened 2024-04-04
fixing CMLResultsReader.py - in particular implementing the workaroun…

maciekswat:
…d for SWIG-generated map<int, str> wrapper inside extract_cell_type_cpp_info_from_file_name. Addresses https://github.com/CompuCell3D/CompuCell3D/issues/712

Fixing user-reported error


fyffep (approved):
lgtm

PR #716, opened 2024-04-28
added updated settings files that include settings used by Demos Panel.

maciekswat:
Added missing settings DemoPath and last time it was changed


fyffep (approved):
Thanks. These files work on my machine.

fyffep:
Although these files work, I believe that the `LastDemoEditTime` setting should be set to 0. The DemoBrowser will only generate an index file if the last index was missing or outdated, and using a value of 0 will ensure that this always happens when we release.
https://github.com/CompuCell3D/cc3d-player5/blob/d006315d85336ec412cbde3315c098bc767570bc/cc3d/player5/Plugins/ViewManagerPlugins/DemoBrowser.py#L167

tjsego (approved):

maciekswat:
good point @fyffep . Will update those

maciekswat:
Fixed. Merging

PR #717, opened 2024-05-03
Fixed/added Python API screenshot features

tjsego:
* Makes visualization synchronous by default, to avoid potential race condition with tracking fields. 
* Adds programmatic access to screenshot data. 
* Adds API features to programmatically set screenshot data. 
* Adds generation of screenshot data from Python API visualization frames.

### DIFF
--- a/cc3d/CompuCellSetup/CC3DPy.py
+++ b/cc3d/CompuCellSetup/CC3DPy.py
@@ -1,3 +1,4 @@
+import os
 import time
 from cc3d import CompuCellSetup
 from cc3d.core.GraphicsUtils.CC3DPyGraphicsFrame import CC3DPyGraphicsFrameClient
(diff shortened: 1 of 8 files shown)


cc3d/CompuCellSetup/CC3DPy.py
```diff
@@ -127,6 +128,85 @@ def visualization():
...
+        :param _fp: absolute path of saved data
+        :return: true is loaded
+        """
+
+        if not os.path.isfile(_fp):
+            raise FileNotFoundError('Screenshot data file does not exist at path', _fp)
+
+        sm = CompuCellSetup.persistent_globals.screenshot_manager
+        if sm is None:
+            raise RuntimeError('Screenshot manager not available at this time')
```
fyffep, line 185: Since these lines are used 4 times, it could be extracted into a private `get` method that throws the error.

fyffep, line 185: nit: Since these lines are used 4 times, it could be extracted into a private get method that throws the error.

cc3d/core/GraphicsUtils/CC3DPyGraphicsFrame.py
```diff
@@ -5,6 +5,8 @@
 # todo: add CC3D logo to window title icon
 # todo: disable built-in key commands for closing render windows
 # todo: add support for additional plots (e.g., tracking fields)
+# todo: improve synchronization to allow asynchronous execution and visualization while handling race conditions
+# todo: implement setting a frame by provided screenshot data
```
fyffep, line 9: There is a quote from Bob Martin that I love because I am very guilty of it. It goes, "TODO means 'Don't do.'" We lose them or forget them when they're in the code. Could you create Issues for these?

cc3d/core/GraphicsUtils/CC3DPyGraphicsFrameIO.py
```diff
@@ -470,3 +470,30 @@ def control(self, control):
...
+
+    def __init__(self):
+
+        super().__init__()
+
+        self.screenshot_data = None
+
+    def process(self, proc):
+        """Get screenshot data"""
+        self.screenshot_data = proc.frame.get_screenshot_data().to_json()
```
fyffep, line 486: Is there any chance of a `null` screenshot dict? I don't think so, but I want to check with you.

fyffep (approved):

fyffep:
Looks great. However, I did not test it. Can you provide some brief testing instructions?

maciekswat (approved):
Looks good. The only comment I have is to move static method to separate module e.g. ScreenshotUtils.py (we can do it later, or as @fyffep suggests never - following his TODO comment). In general static method suggests that it is safe to extract it from a class and move to a module where they can be reused outside of the context of the class. This is only a matter of my preference and not a strict requirement. In your case it probably makes sense to have then inside the class given the context but often you see a general purpose methods that are made static and belong to a class whereas in fact they would be much better off if they were extracted. Enough ranting...

PR #719, opened 2024-05-11
Feature/3 d mode toggle

maciekswat:
This is the PR that accompanies https://github.com/CompuCell3D/cc3d-player5/pull/61


cc3d/CompuCellSetup/simulation_setup.py
```diff
@@ -341,6 +343,14 @@ def initialize_field_extractor_objects():
...
 
     persistent_globals = CompuCellSetup.persistent_globals
+
+    try:
+        # do not reinitialize storage and extractor if they already exist
+        _ = persistent_globals.persistent_holder['field_storage']
+        return
+    except KeyError:
+        pass
+
```
fyffep, line 353: Can this be refactored into the following?
```
# Handle case where persistent_holder is null
if not hasattr(persistent_globals, 'persistent_holder'):
    persistent_holder = {}

# Do not reinitialize storage and extractor if they already exist
if 'field_storage' in persistent_globals.persistent_holder:
    return
```

fyffep (changes requested):

cc3d/CompuCellSetup/simulation_setup.py
maciekswat: replaced it with 
```python
    if 'field_storage' in persistent_globals.persistent_holder.keys():
        return
```

but did it in the `feature/ocl_fixes_new` branch

https://github.com/CompuCell3D/CompuCell3D/pull/721

fyffep: Works for me

fyffep (approved):

tjsego (approved):

PR #721, opened 2024-05-26
Feature/ocl fixes new

maciekswat:
Adds OpenCL out-of-the-box integration + fixes installation of DLLs in non-standalone mode


CompuCell3D/CMakeLists.txt
```diff
@@ -45,7 +45,12 @@ OPTION(FLEX_PYTHON_SPECS "Allow Flexible Python Path specification" OFF)
 
 
 find_package(Python3 REQUIRED COMPONENTS Interpreter Development NumPy)
-
+if(${CMAKE_SYSTEM_NAME} STREQUAL Windows)
+    get_filename_component(PYTHON_BASE_DIR ${Python3_EXECUTABLE} DIRECTORY)
+else(${CMAKE_SYSTEM_NAME} STREQUAL Windows)
+    get_filename_component(PYTHON_BASE_DIR ${Python3_EXECUTABLE} DIRECTORY)
+    get_filename_component(PYTHON_BASE_DIR ${PYTHON_BASE_DIR} DIRECTORY)
+endif()
```
fyffep, line 53: Recommended code:
```
if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
    get_filename_component(PYTHON_BASE_DIR "${Python3_EXECUTABLE}" DIRECTORY)
else()
    get_filename_component(PYTHON_BASE_DIR "${Python3_EXECUTABLE}" DIRECTORY)
    get_filename_component(PYTHON_BASE_DIR "${PYTHON_BASE_DIR}" DIRECTORY)
endif()
```
1. this eliminates the redundant condition
2. adding quotes ensures that directories containing spaces can be handled

CompuCell3D/CMakeLists.txt
```diff
@@ -148,10 +159,26 @@ if(BUILD_STANDALONE)
...
+    if(${CMAKE_SYSTEM_NAME} STREQUAL Windows)
+        SET(COMPUCELL3D_RUNTIME_DESTINATION ${PYTHON_BASE_DIR}/Library/bin)
+        SET(COMPUCELL3D_LIBRARY_DESTINATION ${PYTHON_BASE_DIR}/Library/bin)
+        SET(COMPUCELL3D_ARCHIVE_DESTINATION ${PYTHON_BASE_DIR}/Library/bin)
+    else(${CMAKE_SYSTEM_NAME} STREQUAL Windows)
+        SET(COMPUCELL3D_RUNTIME_DESTINATION ${PYTHON_BASE_DIR}/lib)
+        SET(COMPUCELL3D_LIBRARY_DESTINATION ${PYTHON_BASE_DIR}/lib)
+        SET(COMPUCELL3D_ARCHIVE_DESTINATION ${PYTHON_BASE_DIR}/lib)
+
+    endif()
```
fyffep, line 180: Same as above. Remove the redundant condition and add quotes.

CompuCell3D/core/CompuCell3D/steppables/PDESolvers/OpenCL/DiffusionSolverFE_OpenCL.cpp
```diff
@@ -740,9 +740,23 @@ void DiffusionSolverFE_OpenCL::initImpl() {
 
     //    "lib/CompuCell3DSteppables/OpenCL/DiffusionKernel.cl" };
 
+//    const char *kernelSource[] = {
+//            (string("lib/site-packages/cc3d/cpp/CompuCell3DSteppables/OpenCL/GPUSolverParams.h")).c_str(),
+//            (string("lib/site-packages/cc3d/cpp/CompuCell3DSteppables/OpenCL/DiffusionKernel.cl")).c_str()};
+//
+
+    string cc3d_open_cl_solvers_dir = string(getenv("CC3D_OPENCL_SOLVERS_DIR"));
```
fyffep, line 748: Should you handle the case where `getenv` returns null? Also, should you abort the diffusion when the directory does not exist?

fyffep (changes requested):

CompuCell3D/CMakeLists.txt
maciekswat: Fixed

CompuCell3D/core/CompuCell3D/steppables/PDESolvers/OpenCL/DiffusionSolverFE_OpenCL.cpp
maciekswat: Fixed: Added handling of null pointer but when directory does not exist we get notification from OpenCL that it cannot find required file which is good enough for now

CompuCell3D/CMakeLists.txt
maciekswat: Fixed

fyffep: I see that the redundant condition was removed, but I do not see the quotes. Did you have a strong preference for one way over another?

CompuCell3D/core/CompuCell3D/steppables/PDESolvers/OpenCL/DiffusionSolverFE_OpenCL.cpp
fyffep: Nice!

fyffep:
Except for the my last comment, I'd say this PR is ready for approval.

tjsego (commented):
Overall, looks good. My only overarching comment is about whether we should integrate OpenCL support with the [khronos-opencl-icd-loader](https://anaconda.org/conda-forge/khronos-opencl-icd-loader) package and, if so, whether such support should be accomplished here or in a future version. Long-term, this could allow us to distribute a second `cc3d` conda package (e.g., `cc3d-opencl`) with built-in OpenCL support.

maciekswat:
This is a good point and we can do it in the future (after testing). This PR aims to re-enable OpenCL support for Solver because what we have right now does not work. This PR fixes it. Let's roll it in and we can do the next round of revisions in the next versions of CC3D

CompuCell3D/CMakeLists.txt
maciekswat: this is resolved although I don't know why GitHub does not update it here:
https://github.com/CompuCell3D/CompuCell3D/blob/feature/ocl_fixes_new/CompuCell3D/CMakeLists.txt

fyffep: Nice, thank you for the new commit (a7fc6fafb94cb7cf251f8825fd48e87cae07e6b5). I still see a few places in this file where we might need quotes to handle directories, but I do not wish to be too pedantic. Your choice to fix them. 
PR approved.

fyffep (approved):

tjsego (approved):

PR #727, opened 2024-06-06
Restored python backward compatibility

tjsego:
Might be more, but these are what I found. But in general, there's no reason to pin to Python 3.10+.


maciekswat (approved):
looks good

PR #752, opened 2024-08-21
Minor improvement map handling.

tjsego:
Addresses intermittent bugs related to XML data handling through the Python API.

### DIFF
--- a/cc3d/core/XMLUtils.py
+++ b/cc3d/core/XMLUtils.py
@@ -3,25 +3,31 @@ from cc3d.cpp import CC3DXML
 from cc3d.cpp.CC3DXML import *
 
 
-def dictionaryToMapStrStr(_dictionary:dict)->object:
+def dictionaryToMapStrStr(_dictionary: dict) -> object:
     """
-    converts python dictionaty to c++ map (std::map<std:string, std:string>)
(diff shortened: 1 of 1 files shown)


cc3d/core/XMLUtils.py
```diff
@@ -3,25 +3,31 @@
...
+        return map_str_str
+    except TypeError:
+        return CC3DXML.MapStrStr({str(k): str(v) for k, v in _dictionary.items()})
 
 
 class ElementCC3D:
-    def __init__(self, _name, _attributes={}, _cdata=""):
+    def __init__(self, _name, _attributes: dict = None, _cdata=""):
+        if _attributes is None:
+            _attributes = {}
```
fyffep, line 24: Why use a default value of None?

fyffep (approved):

cc3d/core/XMLUtils.py
tjsego: In general, it's good practice to avoid using mutable objects for default values. See [here](https://docs.python-guide.org/writing/gotchas/) for details.

fyffep: Nice find! I wasn't aware of that nuance.

maciekswat (approved):
Looks good.

PR #756, opened 2024-08-31
Feature/elongated cells tutorial

maciekswat:
This PR is a companion for Reference manual PR: https://github.com/CompuCell3D/CompuCell3DReferenceManual/pull/30

It contains demos that are referenced in the new tutorial


fyffep:
Few notes for ease of use:
1. Could you reduce this to 3-4 demos, depending on how much duplicate code there is? I would find it difficult to use these demos if the differences between each one are minute. 
2. Could you add a README.txt file in the same directory as each .cc3d file? Each one should have a sentence or two to explain what makes the tutorial unique, but the rest can be copy-pasted between the demos. 
3. Can you rename the .cc3d files so that each one is unique? Even names like `ElongatedCellsTutorial10.cc3d` are fine. It's a minor nitpick, but having a unique name for each demo makes it easier to find in the Demo Browser (and perhaps Windows Explorer).

tjsego:
Need to change target develop.

maciekswat:
@fyffep 

"Could you reduce this to 3-4 demos, depending on how much duplicate code there is? I would find it difficult to use these demos if the differences between each one are minute."

Normally, I would use fewer demos but we are lacking a really step-by-step tutorial that showcases how to go from trivial simulation to not so trivial. I also wanted to have a demo for every increment in model sophistication so that new users do not have to second-guess which mechanisms were added.  With your Real-World examples, we will have a nice set of tutorials - this one with more simulations and the once you develop that are targeted for slightly more advanced users.

maciekswat:
@fyffep I have just added README's and renamed the files

CompuCell3D/core/Demos/CompuCellPythonTutorial/ElongatedCellsTutorial/Tutorial_09/README.txt
```diff
@@ -0,0 +1 @@
+This demo shows how to create in Python multiple cells clusters  We are also applying a random force to the cells that "cap" the cluster
```
fyffep, line 1: nit: Add punctuation
It may also be worth mentioning that this is part 9 of 10 for the Elongated Cells walk through, in case it is not already obvious by the name.

fyffep (approved):
I was not sure at first if each demo was part of the step-by-step guide in the Reference Manual. Now that I know, I approve.

tjsego (approved):
Looks good to me.

PR #759, opened 2024-09-03
Added ReactionDiffusionModels from Gülsemay YİĞİT

fyffep:
Full credit to Gülsemay YİĞİT for these models. I tested each one of them to ensure that they run on my machine since they were created on Mac OS.


maciekswat (approved):
Excellent! Many thanks for putting it together @fyffep and even bigger thanks to Gülsemay YİĞİT

PR #762, opened 2024-09-08
Fixed a random segmentation fault that occurred when end-cell, part o…

maciekswat:
…f a cluster connected via curvature links disappeared. The code tried to establish links between neighbors of the disappearing cells, and it assumed that there were two neighbors, but in reality for end-cell there is only one. I turned off this code. If the "middle" cell disappears, we do not try to link two fragments together. I think this makes more sense

Here is the illustration:
image
When the green cell disappears the code tries to establish a link between two neighbors but in reality, there is only one link. This led to seg fault.


fyffep:
Do the relevant demos still have the same behavior? It seems like this is a quick fix to move a very broken feature into a slightly less broken feature.

tjsego (approved):

maciekswat:
@fyffep this is a quick fix but I believe that it is a correct fix. if the three-cell cluster breaks because the middle one o disappears there is no reason why the two remaining ones should remain connected just because the "middleman" disappeared. Keep in mind that there is a mechanism for the two to reconnect but they have to physically touch each other again, but what if during the disappearance of the middle cell, the two remaining ones are far apart? They should not be connected.

fyffep:
Thanks for the explanation

PR #764, opened 2024-09-23
Fixed cell field extraction argument handling

tjsego:
Some usages of `extract_cell_field_data` pass `None` for shell optimization, particularly related to handling `ScreenshotData` instances.


maciekswat (approved):
Look OK. just out of curiosity, when would cell_shell optimization be `None`?

PR #769, opened 2024-10-28
Feature/py surface tracker

tjsego:
* Adds [SurfaceTrackerPlugin](https://github.com/CompuCell3D/CompuCell3D/blob/master/CompuCell3D/core/CompuCell3D/plugins/SurfaceTracker/SurfaceTrackerPlugin.h) to Python API. 
* Adds steerable interface to Python API ContactPlugin. 
* Expands support for Python API SurfacePlugin (*e.g.*, neighborhoods). 
* Minor bug fixes and cleanup in build.

### DIFF
--- a/CompuCell3D/core/CompuCell3D/CMakeLists.txt
+++ b/CompuCell3D/core/CompuCell3D/CMakeLists.txt
@@ -56,12 +56,6 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL Windows)
       TARGET_INCLUDE_DIRECTORIES(CompuCell3DCML PUBLIC ${INCLUDE_DIRS})
       TARGET_LINK_LIBRARIES(CompuCell3DCML cc3d::CompuCellLib cc3d::XMLParserExpatLib)
       INSTALL_TARGETS(/bin CompuCell3DCML)
-  else(BUILD_CPP_ONLY_EXECUTABLE)
-
-      ADD_EXECUTABLE(mainSimpleDemo mainSimpleDemo.cpp)
-      TARGET_INCLUDE_DIRECTORIES(mainSimpleDemo PUBLIC ${INCLUDE_DIRS})
-      TARGET_LINK_LIBRARIES(mainSimpleDemo cc3d::CompuCellLib)
-      INSTALL_TARGETS(/bin mainSimpleDemo)
 
   endif(BUILD_CPP_ONLY_EXECUTABLE)
 
--- a/CompuCell3D/core/CompuCell3D/Potts3D/Potts3D.cpp
+++ b/CompuCell3D/core/CompuCell3D/Potts3D/Potts3D.cpp
@@ -1381,12 +1381,14 @@ void Potts3D::update(CC3DXMLElement *_xmlData, bool _fullInitFlag) {
 
             //we do not allow steering for motility specified by type
             sim->ppdCC3DPtr->temperature = _xmlData->getFirstElement("FluctuationAmplitude")->getDouble();
+            temperature = _xmlData->getFirstElement("FluctuationAmplitude")->getDouble();
             fluctAmplGlobalReadFlag = true;
         }
     }
 
     if (!fluctAmplGlobalReadFlag && _xmlData->getFirstElement("Temperature")) {
         sim->ppdCC3DPtr->temperature = _xmlData->getFirstElement("Temperature")->getDouble();
+        temperature = _xmlData->getFirstElement("Temperature")->getDouble();
     }
 
     if (_xmlData->getFirstElement("RandomSeed")) {
--- a/cc3d/core/PyCoreSpecs.py
+++ b/cc3d/core/PyCoreSpecs.py
@@ -5379,6 +5379,114 @@ class MomentOfInertiaPlugin(_PyCorePluginSpecs):
         return cls()
 
 
+class SurfaceTrackerPlugin(_PyCorePluginSpecs, _PyCoreSteerableInterface):
+    """ SurfaceTracker Plugin """
+
+    name = "surface_tracker"
+    registered_name = "SurfaceTracker"
+
+    def __init__(self,
+                 neighbor_order: int = None,
+                 max_neighbor_order: int = None,
+                 max_neighbor_distance: float = None):
+
+        super().__init__()
+
+        self.max_neighbor_distance = max_neighbor_distance
+        self.max_neighbor_order = max_neighbor_order
+        self.neighbor_order = neighbor_order
+
+    @property
+    def xml(self) -> ElementCC3D:
+        """
+        CC3DXML element; generated from specification dictionary
+
+        :return: CC3DML XML element
+        """
+        self._el = self.generate_header()
+        if self.neighbor_order is not None:
+            self._el.ElementCC3D('NeighborOrder', {}, str(self.neighbor_order))
+        elif self.max_neighbor_order is not None:
+            self._el.ElementCC3D('MaxNeighborOrder', {}, str(self.max_neighbor_order))
+        elif self.max_neighbor_distance is not None:
+            self._el.ElementCC3D('MaxNeighborDistance', {}, str(self.max_neighbor_distance))
+        return self._el
+
+    @classmethod
+    def from_xml(cls, _xml: CC3DXMLElement):
+        """
+        Instantiate an instance from a CC3DXMLElement parent instance
+
+        :param _xml: parent xml
+        :return: python class instace
+        :rtype: MomentOfInertiaPlugin
+        """
+        el = cls.find_xml_by_attr(_xml)
+
+        neighbor_order = None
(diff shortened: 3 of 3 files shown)


CompuCell3D/core/CompuCell3D/Potts3D/Potts3D.cpp
```diff
@@ -1381,12 +1381,14 @@ void Potts3D::update(CC3DXMLElement *_xmlData, bool _fullInitFlag) {
...
             //we do not allow steering for motility specified by type
             sim->ppdCC3DPtr->temperature = _xmlData->getFirstElement("FluctuationAmplitude")->getDouble();
+            temperature = _xmlData->getFirstElement("FluctuationAmplitude")->getDouble();
             fluctAmplGlobalReadFlag = true;
         }
     }
 
     if (!fluctAmplGlobalReadFlag && _xmlData->getFirstElement("Temperature")) {
         sim->ppdCC3DPtr->temperature = _xmlData->getFirstElement("Temperature")->getDouble();
+        temperature = _xmlData->getFirstElement("Temperature")->getDouble();
```
fyffep, line 1391: Why do we need two variables for temperature? We could use a comment for this.

CompuCell3D/core/CompuCell3D/Potts3D/Potts3D.cpp
```diff
@@ -1381,12 +1381,14 @@ void Potts3D::update(CC3DXMLElement *_xmlData, bool _fullInitFlag) {
 
             //we do not allow steering for motility specified by type
```
fyffep, line 1382: Since steering is not allowed, it's odd that this code is in `update()`. However, I would avoid a refactor.

cc3d/core/PyCoreSpecs.py
```diff
@@ -1543,7 +1543,7 @@ def xml(self) -> ElementCC3D:
                 fa_el.ElementCC3D("FluctuationAmplitudeParameters", {"CellType": _type,
                                                                      "FluctuationAmplitude": str(_fa)})
         else:
-            self._el.ElementCC3D("FluctuationAmplitude", {}, str(self.spec_dict["fluctuation_amplitude"]))
+            self._el.ElementCC3D("Temperature", {}, str(self.spec_dict["fluctuation_amplitude"]))
```
fyffep, line 1546: If neither FluctuationAmplitude nor Temperature is set, we default to 0.0, right?

cc3d/core/PyCoreSpecs.py
```diff
@@ -3116,12 +3157,19 @@ class ContactPlugin(_PyCorePluginSpecs):
     name = "contact"
     registered_name = "Contact"
 
-    check_dict = {"neighbor_order": (lambda x: x < 1, "Neighbor order must be positive")}
+    check_dict = {"neighbor_order": (lambda x: x is not None and x < 1, "Neighbor order must be positive"),
+                  "depth": (lambda x: x is not None and x < 1, "Depth must be positive")}
```
fyffep, line 3161: I am not sure about the `check_dict` approach. By setting `self.check_dict` rather than passing in check_dict to `self.check_inputs()`, we increase the potential for race conditions (if any of this is multi-threaded). It would be simple & concise to use if-else statements as opposed to this method.

cc3d/core/PyCoreSpecs.py
```diff
@@ -5309,6 +5379,114 @@ def from_xml(cls, _xml: CC3DXMLElement):
...
+
+        :param _xml: parent xml
+        :return: python class instace
+        :rtype: MomentOfInertiaPlugin
+        """
+        el = cls.find_xml_by_attr(_xml)
+
+        neighbor_order = None
+        max_neighbor_order = None
+        max_neighbor_distance = None
```
fyffep, line 5428: What are the default values if these are not specified in the XML?

cc3d/core/PyCoreSpecs.py
```diff
@@ -5309,6 +5379,114 @@ def from_xml(cls, _xml: CC3DXMLElement):
...
+    @neighbor_order.setter
+    def neighbor_order(self, _val: Optional[int]):
+        if _val is None:
+            val = _val
+        else:
+            val = int(_val)
+            if val <= 0:
+                raise ValueError('Value must be positive')
+            self.max_neighbor_order = None
+            self.max_neighbor_distance = None
```
fyffep, line 5451: I am not sure about having these side effects wrapped in the setters. Is it normal for this plugin?

CompuCell3D/core/CompuCell3D/Potts3D/Potts3D.cpp
tjsego: Probably one for @maciekswat to address. This one pre-dates me.

CompuCell3D/core/CompuCell3D/Potts3D/Potts3D.cpp
tjsego: Unsure whether this rule still applies. @maciekswat ?

cc3d/core/PyCoreSpecs.py
tjsego: In the Python API, a value is always set. The default value is 10 as a default argument, per recommendations in the code snippets.

cc3d/core/PyCoreSpecs.py
tjsego: Since `check_dict` is a class attribute, race conditions shouldn't be a problem. Did you have a specific scenario in mind where a race condition might occur?

cc3d/core/PyCoreSpecs.py
tjsego: `None` values here correspond to no specification in the XML. Each has their own effect, but what's defined here should be consistent with analogous XML spec.

cc3d/core/PyCoreSpecs.py
tjsego: This is imposing object state consistently with how the XML spec is parsed and implemented, to avoid confusion for the user. For example, if two underlying attributes are not `None`, the XML will only use one of them, so the object states reflect this.

cc3d/core/PyCoreSpecs.py
fyffep: Ah, you are right. At first, I read it as static.

cc3d/core/PyCoreSpecs.py
fyffep: That makes sense. I like that design, then.

fyffep (approved):

CompuCell3D/core/CompuCell3D/Potts3D/Potts3D.cpp
maciekswat: Steering is allowed. Did we disallow it?

cc3d/core/PyCoreSpecs.py
```diff
@@ -1543,7 +1543,7 @@ def xml(self) -> ElementCC3D:
                 fa_el.ElementCC3D("FluctuationAmplitudeParameters", {"CellType": _type,
                                                                      "FluctuationAmplitude": str(_fa)})
         else:
-            self._el.ElementCC3D("FluctuationAmplitude", {}, str(self.spec_dict["fluctuation_amplitude"]))
```
maciekswat, line 1546: Can we have both? I think current Potts has both Temperature and FluctuationAmplitude and internally both of those tags are mapped to temperature variable

CompuCell3D/core/CompuCell3D/CMakeLists.txt
```diff
@@ -56,12 +56,6 @@ if(${CMAKE_SYSTEM_NAME} STREQUAL Windows)
       TARGET_INCLUDE_DIRECTORIES(CompuCell3DCML PUBLIC ${INCLUDE_DIRS})
       TARGET_LINK_LIBRARIES(CompuCell3DCML cc3d::CompuCellLib cc3d::XMLParserExpatLib)
       INSTALL_TARGETS(/bin CompuCell3DCML)
-  else(BUILD_CPP_ONLY_EXECUTABLE)
```
maciekswat, line 59: This is a good move. We are not really using it anymore

maciekswat (changes requested):
Looks good overall - let's resolve this FluctuationAmplitude question but otherwise we shold be ready to merge it

CompuCell3D/core/CompuCell3D/Potts3D/Potts3D.cpp
fyffep: @maciekswat

maciekswat: This was introduced when we wanted to move away from temperaturature

maciekswat (approved):
looks good to me

Issue #770, opened 2024-10-31
Python API: Issue using ConnectivityPlugin


LVeschini78:
### Provide environment information

```bash
OS: Darwin
Platform: Intel64
CC3D version: 4.6.0
Install type: Conda Install on Python 3.10
```

### Provide Simulation files (as a zipped archive) and ascreenshots

[WS_Example_code.zip](https://github.com/user-attachments/files/17587220/WS_Example_code.zip)
![CP_Crash_example](https://github.com/user-attachments/assets/3f1f92ec-1173-42eb-b69f-fc80821a5298)
![CGP_Successful_run_example](https://github.com/user-attachments/assets/bcb49c9f-91be-46d1-82c9-712f60c1a734)

### Installation Method

via conda package

### Which area(s) are affected? (Select all that apply)

Documentation, Runtime

### Steps To Reproduce

- Download the example code and change extension to .ipynb (it is not a real zipfile, just changed the extension to allow upload)
- The Notebook uses an example extracted from the latest CC3D WS, the angiogenesis simulation (no particular reason for choosing that, just the shortest simulation I had at hand).
- Run the first cell (just imports)
- Focus on lines 89-97 in the second cell, the AngioSim class definition.
- Run the simulation (3rd cell) with either CP or CGP to get the results in example screenshots.

### Current vs. Expected behavior

# Context
In any cc3d simulation:

    <Plugin Name="Connectivity">
        <Penalty>10000000</Penalty>
    </Plugin>

Omitting `<Penalty>10000000</Penalty>` will crash CC3D player.

# Within the CC3D Py API

## Issue
As per included examples a cc3d simservice running with the spec below will crash.

        simConnectivity = ConnectivityPlugin()
        self.specs.append(simConnectivity)

It seems there is no immediate way to pass the "Penalty" parameter according to class docstrings and documentation.

## Solution
Using ConnectivityGlobalPlugin with fast enabled, runs the simulations withouth problems.

        simConnectivity = ConnectivityGlobalPlugin(fast=True)
        self.specs.append(simConnectivity)

### Note!
According to documentation https://compucell3dreferencemanual.readthedocs.io/en/latest/connectivity.html
The ConnectivityGlobalPlugin also require to turn connectivity "on" in a steppable. 
For example:

    class ConnectivityElongationSteppable(SteppableBasePy):
        def __init__(self,_simulator,_frequency=10):
            SteppableBasePy.__init__(self,_simulator,_frequency)

[...]

PR #777, opened 2025-01-25
Removed Penalty param from ConnectivityPlugin

fyffep:
Resolves [770](https://github.com/CompuCell3D/CompuCell3D/issues/770).
Since the value of the Penalty parameter has no effect on this plugin as long as it is a positive number, I have removed it. We will also need to update the docs to explain that Penalty no longer has a function. 
https://compucell3dreferencemanual.readthedocs.io/en/latest/connectivity.html

I left the ConnectivityLocalFlexPlugin alone since it is deprecated. This plugin uses a ConnectivityStrength (accessible in Python) to control the energy returned, which, in turn, controls whether or not the pixel copies can occur.


maciekswat:
@fyffep  This looks good to me. Did you check the demos to see if they behave the same as before this fix? I think they should because there is nothing in this PR that would say otherwise, but if you could check it, that would be great.

maciekswat:
@fyffep could you change the target for this PR to go to the develop branch?

fyffep:
> @fyffep could you change the target for this PR to go to the develop branch?

I changed it. There are a few commits related to README changes that are in master but not develop:
https://github.com/CompuCell3D/cc3d-player5/commits/master
image

fyffep:
I tested this feature with the below project. I observed that it's just an ordinary cell blob. 
[connectivityTestDebugging2025-01.zip](https://github.com/user-attachments/files/19822899/connectivityTestDebugging2025-01.zip)

Sorry about the discrepancy between branches. The only part you need to care about is this commit, and the rest of the files are just from merging commits that were mistakenly pushed to master instead of develop. 
https://github.com/CompuCell3D/CompuCell3D/pull/777/commits/bb157889d5b3d8a8a9d17678ff63635aff277a0c

tjsego (approved):
Looks good.

maciekswat (approved):
looks good

PR #821, opened 2026-03-01
Release/4.8.0

maciekswat:
Most of the changes have been reviewed. This is release 4.8.0 PR. There are a few newer commits, though


cc3d/CompuCellSetup/simulation_utils.py
```diff
@@ -164,3 +166,32 @@ def val_check(inv_val_str):
     return container_int
 
 
+def extract_error_block(tb: str):
```
fyffep, line 169: Can you put this behind a setting? A "DebugMode" setting would be useful for both us programmers and users alike. There will always be times when a truncated error message will make one wonder, "Am I missing something because I'm not seeing the whole error?"

fyffep: The other commits until this point look good. I started reviewing at commit 2e42456fc79d698e4206df8cbb0d61f44c6c034 ("Introduced automated way to handle version , build number, revision a…")

maciekswat: you still see the entire error. What this function does is that it extracts entire exception text when it spans multi lines - previously we extracted just the first line and it did not give the user goo picture of went wrong. Here ,we are 4adding a convenience function that extracts a full exception text and then we combine it with a full tracebac. So the behavior is the the same as before except the user gets to see full exception text, instead of just a first line. So both users and developers get to see all information they need

fyffep: Okay. One more thing, can you wrap extract_error_block with a try-catch? We want to avoid showing any errors that could occur in that function if the real error was elsewhere. 
Otherwise, looks good.

fyffep (approved):

PR #827, opened 2026-04-18
Added support for in-memory cache for settings. this way get and set …

maciekswat:
…operations are instantaneous - no disks write while the simulation is running, and we write updated settings at the end of the simulation. 

Some users reported issues with the settings database access. So this fix should help a lot


bartjuw (approved):
Looks good

PR #832, opened 2026-06-28
Feature/steering fix

maciekswat:
Fixing steering code - part of it (MEtadata handling) was leading to runtime crashes


fyffep (approved):

CompuCell3D/CMakeLists.txt
bartjuw, line 1: The version number for CMakelists id 4.10., while _version.py is 4.9.0, is it ok to have divergent version numbers here, or is the _version.py auto updated at some point?

bartjuw (approved):
Looks good, only issue I saw was version numbering. If it is expected then changes fine.

PR #833, opened 2026-07-04
Feature/4.10.0 fix omp ocl merge

maciekswat:
This PR adds OPenCL PDE solvers that work "out of the box" no compilation required. Just making sure that you have correct drivers on your machine. They work on OSX, WIndows and Linux


CompuCell3D/core/PublicUtilities/ParallelUtilsOpenMP.cpp
```diff
@@ -167,13 +167,36 @@ void ParallelUtilsOpenMP::destroyLock(OpenMPLock_t * _lock){
...
+	if (!_lock || !_lock->nativeLock) {
+		return;
+	}
+
+	omp_destroy_lock(static_cast<omp_lock_t *>(_lock->nativeLock));
+	delete static_cast<omp_lock_t *>(_lock->nativeLock);
+	_lock->nativeLock = 0;
+}
+void ParallelUtilsOpenMP::setLockV1(OpenMPLockV1_t * _lock){
+	omp_set_lock(static_cast<omp_lock_t *>(_lock->nativeLock));
```
fyffep, line 197: 1. Considering assert that _lock is not null with `assert(_lock);`
2. These methods should due nothing if OpenCL is not loaded properly, right? Are there any safeguards for that?

CompuCell3D/core/PublicUtilities/ParallelUtilsOpenMP.cpp
```diff
@@ -167,13 +167,36 @@ void ParallelUtilsOpenMP::destroyLock(OpenMPLock_t * _lock){
...
-	setLock(&pyWrapperGlobalLock);
-}
+void ParallelUtilsOpenMP::unsetLock(OpenMPLock_t * _lock){
+	omp_unset_lock(_lock);
+}
+
+void ParallelUtilsOpenMP::initLockV1(OpenMPLockV1_t * _lock){
+	if (!_lock->nativeLock) {
+		_lock->nativeLock = new omp_lock_t;
+	}
```
fyffep, line 179: GPT complains that this is not idempotent or thread-safe. Creating a lock that is already initialized will throw a non-specific error.

CompuCell3D/core/PublicUtilities/ParallelUtilsOpenMP.cpp
```diff
@@ -167,13 +167,36 @@ void ParallelUtilsOpenMP::destroyLock(OpenMPLock_t * _lock){
...
+
+	omp_init_lock(static_cast<omp_lock_t *>(_lock->nativeLock));
+}
+void ParallelUtilsOpenMP::destroyLockV1(OpenMPLockV1_t * _lock){
+	if (!_lock || !_lock->nativeLock) {
+		return;
+	}
+
+	omp_destroy_lock(static_cast<omp_lock_t *>(_lock->nativeLock));
+	delete static_cast<omp_lock_t *>(_lock->nativeLock);
```
fyffep, line 191: Further, GPT also recommends using a single struct pointer to hold the _lock. This way, there is no heap allocation. You can just set and unset the `nativeLock` value. IMO, this conserves a trivial amount of memory.

fyffep (commented):
I had AI review the code; its concerns seem non-blocking. It looks good overall.

CompuCell3D/core/PublicUtilities/ParallelUtilsOpenMP.cpp
maciekswat: Good point on _lock. In current usage this wrapper is only called with an initialized OpenMPLockV1_t, so this is not failing today. This is internal code not public facing API so I would leave it the way it is.
2. is OpenMP-only, not OpenCL, although in order to get OpenCL solvers working we had to introduce OpenMPLockV1_t to prevent mixup between multiple openMp headers - it is too technical to go into details but long story short, OpenMPLockV1_t avoids using  explicit omp_lock_t because this led to a lot of issues with openCL solver.

CompuCell3D/core/PublicUtilities/ParallelUtilsOpenMP.cpp
maciekswat: So this is specifically why OpenMPLockV1_t  was created - to avoid using native lock type explicitly. Also those are tiny objects so heap allocation does not matter much. I would not worry about it.

CompuCell3D/core/PublicUtilities/ParallelUtilsOpenMP.cpp
maciekswat: I have added few asserts to harden the code and prevent obvious segfaults when e.g. *lock is nullptr or trying to reinitialize already initialized lock. . As far as thread safety I would not worry about it much  - this is internal structure that is used in specific context over which we have full control.

maciekswat:
@fyffep take a look at my responses + code hardening and see if these answer the issues you saw in the PR

fyffep (approved):
All of my feedback is addressed. Great work!

CompuCell3D/core/PublicUtilities/ParallelUtilsOpenMP.cpp
fyffep: 1. I see the code change. 👍
2. This sort of makes sense.

